PR #11, opened 2021-10-30
Recommending changes on the Style Guide

rishu-buf:
Suggesting changes to the style guide for better readability and user experience. These recommendations a from my personal POV if these don't align with your goals feel free to avoid/remove them.

Thanks!
Rishu Mehra

### DIFF
--- a/docs/best-practices/style-guide.md
+++ b/docs/best-practices/style-guide.md
@@ -152,4 +152,4 @@ The following recommendations are exceptional style to avoid and set up breaking
 
 * Avoid using nested enums and nested messages. You may end up wanting to use them outside of their context message in the future, even if you do not think so at the moment.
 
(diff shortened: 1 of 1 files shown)


rishu-buf (commented):
What and why I am suggesting the changes?

- Avoid using question marks and WH-words on a Call-to-Action (CTA) as it may be considered that something is missing and the product is not finalized. Although add information using "Note" or "For more information check here" or "refer to [title-to-insert]" these kinds of writing styles.
- Avoid using 'but' in technical or marketing documents as 'but' is used in those situations where you are conveying a negative message.
- Bullet points gives a sentence break and an easy flow of reading. 
- Few general language changes for better understanding and readability.

bufdev:
Thanks for the feedback. There's a few things in here we may incorporate, but we largely respectfully disagree with the calls made here - our word choices in the style guide were intentional. We really do appreciate the effort though!

PR #14, opened 2021-11-04
Add user management section

mfridman:
- Should we omit the section on resource roles and organization base roles? I added a _coming soon_ 
- Consider adding a summary table..

### DIFF
--- a/docs/bsr/user-management.md
+++ b/docs/bsr/user-management.md
@@ -6,6 +6,8 @@ title: User Management
 # User Management
 
 ## Organization Roles
+
+Every user that is part of an organization will have an explicit role. Note that users are unable to modify their own role. If you need to lower your access please have another organization user perform this action, or, leave the organization and request to be re-added with the desired role.
 	
 ### Owner
 
@@ -44,7 +46,7 @@ The most common use-cases for explicit resource roles are:
 - Outside collaborators. This is useful when users outside your organization require access to specific resource(s) within the organization, but you do not want them to be a member of the organization.
 - Elevated permissions for organization members. This is useful when the organization base resource roles are set to `read` and specific user(s) require `write` or `admin` permissions. 
 
-When computing the role on a resource, the highest role will take precedence. Example, an organization has `write` as the base repository role, and the user was granted the `admin` role on a specific repository. The final computed user role on the repository will be `admin`.
+When computing the role on a resource, the highest role will take precedence. For example, an organization has `write` as the base repository role, and the user was granted the `admin` role on a specific repository. The final computed user role on the repository will be `admin`.
 
 ### Owner
 
--- a/docs/bsr/user-management.md
+++ b/docs/bsr/user-management.md
@@ -40,9 +40,11 @@ Organization owners can modify the base resource roles depending on the requirem
 
 > *Resource roles coming soon.*
 
-Resource roles are explicit roles granted to a user on an individual resource, such as a repository, template or plugin.
+Resources such as repositories, templates and plugins are owned by either an individual user or an organization. In the case of user-owned resources, the user is granted the `Owner` role and for organization-owned resources members inherit the [base resource roles](#base-resource-roles) as defined by the organization.
 
-The most common use-cases for explicit resource roles are:
+However, in some situations you'll need to give additional permissions to individual users over a user- or organization-owned resource.
+
+The most common use-cases are:
 
 - Outside collaborators. This is useful when users outside your organization require access to specific resource(s) within the organization, but you do not want them to be a member of the organization.
 - Elevated permissions for organization members. This is useful when the organization base resource roles are set to **Read** and specific user(s) require the **Write** or **Admin** role for specific resource(s). 
--- a/docs/bsr/user-management.md
+++ b/docs/bsr/user-management.md
@@ -44,9 +44,9 @@ Resource roles are explicit roles granted to a user on an individual resource, s
 The most common use-cases for explicit resource roles are:
 
 - Outside collaborators. This is useful when users outside your organization require access to specific resource(s) within the organization, but you do not want them to be a member of the organization.
-- Elevated permissions for organization members. This is useful when the organization base resource roles are set to `read` and specific user(s) require `write` or `admin` permissions. 
+- Elevated permissions for organization members. This is useful when the organization base resource roles are set to **Read** and specific user(s) require **Write** or **Admin** permissions. 
 
-When computing the role on a resource, the highest role will take precedence. For example, an organization has `write` as the base repository role, and the user was granted the `admin` role on a specific repository. The final computed user role on the repository will be `admin`.
+When computing the role on a resource, the highest role will take precedence. For example, an organization has **Write** as the base repository role, and the user was granted the **Admin** role on a specific repository. The final computed user role on the repository will be `admin`.
 
 ### Owner
 
--- a/docs/bsr/user-management.md
+++ b/docs/bsr/user-management.md
@@ -25,7 +25,7 @@ Every user that is part of an organization will have an explicit role. Note that
 
 ### Base resource roles
 
-> *Base resource roles coming soon.*
+> *Configurable base resource roles are coming soon.*
 
 Every organization has a set of base resource roles that apply to all members of the organization. The following are the default roles:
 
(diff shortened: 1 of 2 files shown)


docs/bsr/user-management.md
```diff
@@ -0,0 +1,65 @@
...
+
+> *Base resource roles coming soon.*
+
+Every organization has a set of base resource roles that apply to all members of the organizatio. The following are the default roles:
+
+- Repository: `write`
+- Template: `write`
+- Plugin: `write`
+
+Organization owners can modify the base resource roles depending on the requirements of the organization.
```
johanbrandhorst, line 20: We should put this below the other roles, right? It's a more confusing concept, so maybe lead with the normal roles.

docs/bsr/user-management.md
```diff
@@ -0,0 +1,65 @@
...
+
+### **`Member`**
+
+- Can view resources and users within the organization.
+
+## Resource Roles
+
+> *Resource roles coming soon.*
+
+Resource roles are explicit roles granted to a user on an individual resource, such as a repository, template or plugin.
```
johanbrandhorst, line 40: Link to repository/template/plugin concepts here?

docs/bsr/user-management.md
```diff
@@ -0,0 +1,65 @@
...
+
+The most common use-cases for explicit resource roles are:
+
+- Outside collaborators. This is useful when users outside your organization require access to specific resource(s) within the organization, but you do not want them to be a member of the organization.
+- Elevated permissions for organization members. This is useful when the organization base resource roles are set to `read` and specific user(s) require `write` or `admin` permissions. 
+
+When computing the role on a resource, the highest role will take precedence. Example, an organization has `write` as the base repository role, but the user was mistakingly granted the `read` role on a specific repository. The final computed user role on the repository will be `write`.
```
johanbrandhorst, line 47:
```suggestion
When computing the role on a resource, the highest role will take precedence. For example, an organization has `write` as the base repository role, but the user was mistakingly granted the `read` role on a specific repository. The final computed user role on the repository will be `write`.
```

johanbrandhorst: We could make this sort of thing impossible by checking the base resource role of the organization at invitation time. Github does, we probably should too? Maybe a better example is that an organization admin can be invited to a repository explicitly, but their admin capacity gives them admin on the repository regardless.

docs/bsr/user-management.md
```diff
@@ -0,0 +1,65 @@
...
+
+# User Management
+
+## Organization Roles
+
+### Base resource roles
+
+> *Base resource roles coming soon.*
+
+Every organization has a set of base resource roles that apply to all members of the organizatio. The following are the default roles:
```
johanbrandhorst, line 14:
```suggestion
Every organization has a set of base resource roles that apply to all members of the organization. The following are the default roles:
```

docs/bsr/user-management.md
```diff
@@ -0,0 +1,65 @@
...
+
+When computing the role on a resource, the highest role will take precedence. Example, an organization has `write` as the base repository role, but the user was mistakingly granted the `read` role on a specific repository. The final computed user role on the repository will be `write`.
+
+### **`Owner`**
+
+- Unrestricted access to the resource.
+
+### **`Admin`**
+
+- Can update the resource settings, visibility and deprecation notices.
```
johanbrandhorst, line 55: Visibility updates aren't supported right now, we shouldn't include that here

docs/bsr/user-management.md
mfridman: Good call. I originally had this as an explicit `admin` role on a repository to "override" the base repository `write` role. Will change it back to that example...

docs/bsr/user-management.md
```diff
@@ -0,0 +1,65 @@
...
+
+- Can view resources and users within the organization.
+
+### Base resource roles
+
+> *Base resource roles coming soon.*
+
+Every organization has a set of base resource roles that apply to all members of the organization. The following are the default roles:
+
+| Repository | Template | Plugin |
```
mfridman, line 30: I liked the table view better than the bullet points

CleanShot 2021-11-04 at 14 54 11@2x

johanbrandhorst: Maybe capitalize these so they map to the resource role labels below?

mfridman: Maybe we shouldn't use backticks.. seems unecessary.

CleanShot 2021-11-04 at 14 56 28@2x

johanbrandhorst: Yeah 👍🏻

mfridman: What do you think, does it help with readability?

johanbrandhorst (approved):

docs/bsr/user-management.md
```diff
@@ -0,0 +1,67 @@
...
+- Can delete organization. All resources such as repositories, templates and plugins must be deleted before the organization can be deleted.
+
+### Admin
+
+- Can manage user roles, except owners.
+- Can add and delete resources such as [repositories](../bsr/overview.md#module), [templates](../bsr/remote-generation/concepts/#template) and [plugins](../bsr/remote-generation/concepts/#plugin).
+
+### Member
+
+- Can view resources and users within the organization.
```
SyrieBianco, line 24: Technically, members have write access to resources (as you say below).

It might be better to say something along the lines of

```suggestion
- Can view the organization and its members
- Have the [Base resource role](<LINK TO HEADING HERE>) over the organizations resources, which defaults to [Write](<LINK TO RESOURCE WRITE PERMISSION HERE>).
```

And then in Base resource roles, you can say exactly what you do say.

docs/bsr/user-management.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Base resource roles
+
+> *Base resource roles coming soon.*
```
SyrieBianco, line 28: I would say that *configurable* base resource roles are coming soon.

docs/bsr/user-management.md
```diff
@@ -0,0 +1,67 @@
...
+> *Resource roles coming soon.*
+
+Resource roles are explicit roles granted to a user on an individual resource, such as a repository, template or plugin.
+
+The most common use-cases for explicit resource roles are:
+
+- Outside collaborators. This is useful when users outside your organization require access to specific resource(s) within the organization, but you do not want them to be a member of the organization.
+- Elevated permissions for organization members. This is useful when the organization base resource roles are set to **Read** and specific user(s) require the **Write** or **Admin** role for specific resource(s). 
+
+When computing the role on a resource, the highest role will take precedence. For example, an organization has **Write** as the base repository role, and the user was granted the **Admin** role on a specific repository. The final computed user role on the repository will be **Admin**.
```
SyrieBianco, line 49: Do we want to disambiguate resource roles — which you can be granted by organization membership — from "granting direct access" or "adding outside collaborators"? 

It might be useful to define the ways you can interact with a resource — which is true already for user-owners and any organization members — separate from defining how you'll be able to configure resource-to-user specific roles in the future.

docs/bsr/user-management.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Owner
+
+- Unrestricted access to the resource.
+
+### Admin
+
+- Can update the resource settings and deprecation notices.
+- Can manage resource roles, except owners.
+- Can delete the resource.
```
SyrieBianco, line 59: Can admins delete the resource? This used to be only owners; just confirming that we want that behavior to change. I think I originally thought all destructive behavior (name changing, deletion) would be limited to owners.

docs/bsr/user-management.md
```diff
@@ -0,0 +1,67 @@
...
+
+### Write
+
+- Can perform modifying operations, except for deletes.
```
SyrieBianco, line 63: Might want to be more specific here — this just means like `push`ing and making branches and tags, right? Not updating the resource name, which might also be considered a modifying operation.

Again, I'm not actually sure what's enforced by the new authz, but I'm of the opinion that they should only be able to write to the content of the resource, rather than the resource as a DB entity ¯\_(ツ)_/¯

docs/bsr/user-management.md
mfridman: You're right, deletion should be limited to the owner resource role.

docs/bsr/user-management.md
mfridman: > Might want to be more specific here

Good point, updated with a few concrete examples.

docs/bsr/user-management.md
mfridman: Not sure if this is what you mean, but I added a blurb to this section to expand on the user- and organization-owned resources and the default roles users have .. which is a segue into why we'd give additional permissions to a user over a resource.

Let me know if this is what you were going for?

https://deploy-preview-14--buf.netlify.app/bsr/user-management/#resource-roles

SyrieBianco (approved):

PR #29, opened 2021-12-22
Override key `%s/JAVA_PACKAGE_PREFIX/JAVA_PACKAGE`

doriable:
The correct key for the override for the `java_package` option is `JAVA_PACKAGE`.


timostamm (approved):
TY

PR #31, opened 2022-01-14
Remove coming-soon note for base-resource-roles

seankimdev:
image


SyrieBianco (approved):
Thanks! Would be useful before describing the roles just to drop "These roles are configurable on the organization settings page." But removing the "coming soon" is the important part since it's launched

seankimdev:
> Thanks! Would be useful before describing the roles just to drop "These roles are configurable on the organization settings page." But removing the "coming soon" is the important part since it's launched

Certainly. I added it. Thanks!

PR #35, opened 2022-01-19
Add Image component

lucperkins:
This PR adds an image component to replace the current inline HTML. This component supports captions and also makes images clickable (opening into a fresh tab).

Here's a page that uses this component: https://deploy-preview-35--buf.netlify.app/bsr/documentation.


mfridman (approved):
Was a bit confused how the `src` is handled, but I can see the component is still using the built-in `baseUrl` offered by docusaurus.

Is it possible (do we care) to modify the image size?

lucperkins:
@mfridman Ah, good call. I've added a width property. It's nothing fancy and just uses `width="X%"` but I think that's fairly standard.

mfridman (approved):
lgtm. 

Having one way to compose images and consistency on their presentation is great.

PR #38, opened 2022-01-21
Uncollapse remote generation by default

bufdev:
I think we should advertise this.


lucperkins:
@bufdev The build failures here are caused by a broken link, which I fixed in #39. If you pull latest `main` and push that should fix this.

PR #40, opened 2022-01-24
Buf tour revamp (pt 2)

lucperkins:
- Small changes to breaking doc plus add make start command
- Revamp generate code tour doc
- Revamp BSR login doc


docs/tour/detect-breaking-changes.md
```diff
@@ -46,33 +46,39 @@ For example, change the type of the `Pet.pet_type` field from `PetType` to `stri
...
+---
 pet/v1/pet.proto:1:1:Previously present service "PetStore" was deleted from file.
 pet/v1/pet.proto:20:3:Field "1" on message "Pet" changed type from "enum" to "string".
 pet/v1/pet.proto:44:3:Field "1" with name "pet_id" on message "DeletePetRequest" changed option "json_name" from "petID" to "petId".
 pet/v1/pet.proto:44:10:Field "1" on message "DeletePetRequest" changed name from "petID" to "pet_id".
 ```
 
-> For remote locations that require authentication, see [HTTPS Authentication](../reference/inputs.md#https) and
-> [SSH Authentication](../reference/inputs.md#ssh) for more details.
+:::note
```
mfridman, line 74: We historically avoided the [admonitions](https://docusaurus.io/docs/markdown-features/admonitions) feature to avoid lock-in and iirc there were some styling issues in earlier beta versions. 

The first point is probably irrelevant, but we should see if the admonitions fit nicely within the existing docs. Probably okay, but worth validating the look and feel is what we want.

lucperkins: I agree that the default look and feel isn't super awesome, but we can customize that pretty easily using CSS variables, which I have on my TODO list. In terms of lock-in, I'm personally not super concerned. If we end up re-platforming we can `git grep` to find where we're using it and either create our implementation for the new platform or revert to blockquotes.

lucperkins: Just to get a sense, here are some example blocks with default styling (running locally):

Screen Shot 2022-01-24 at 12 17 48 PM

Agreed that we should customize but that should be pretty trivial, and I do think it's nice to break up the grey-on-black scheme.

mfridman: Great points, I do like how it breaks up the page and draws attention, my only main issue is the default colors just stick out too much .. to the point of distraction. I think once we modify those colors with CSS it'll fit much nicer 👍

lucperkins: Cool cool. I'll submit a follow-up OR with some aesthetic improvements. I've found some nice sets of admonition colors; should be easy to implement.

lucperkins: Okay, follow-up submitted in #41

mfridman, line 74: Afaics we're changing to using admonitions in favor of block quotes for some of these things. We should do a quick scan to make sure we got them all so the styling is consistent, at least for items that are identical like this:

CleanShot 2022-01-27 at 14 54 34@2x

CleanShot 2022-01-27 at 14 54 29@2x

lucperkins: Ah yeah, good call. Let's do blockquotes only for now and reassess later

mfridman (approved):
Looks great, thanks for taking another pass at this. 

Let's be consistent between block quotes or admonitions, but otherwise nice!

PR #41, opened 2022-01-25
Add custom styling for admonition blocks

lucperkins:
This PR adds some custom styling/config for admonition blocks. It's nothing fancy, but I've optimized for subtlety and readability. With the proposed styling:

image

### DIFF
--- a/src/css/custom.css
+++ b/src/css/custom.css
@@ -61,8 +61,10 @@
      */
 
   /* Admonition colors (derived from https://github.com/favoloso/remarkable-admonitions/blob/master/styles/docusaurus-admonitions.css) */
-  --ifm-color-success: rgba(46, 204, 113, 0.1); /* Lime 200 */
+  --ifm-color-success: rgba(46, 204, 113, 0.1);
   --ifm-color-info: rgba(52, 152, 219, 0.1);
+  --ifm-color-danger: rgba(231, 76, 60, 0.1);
+  --ifm-color-warning: rgba(230, 126, 34, 0.1);
 
   --default-text-color: rgb(14, 15, 20);
 }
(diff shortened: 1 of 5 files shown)


src/css/custom.css
```diff
@@ -1,50 +1,56 @@
...
+  /* Custom chevron for the left hand sidebar */
+  --ifm-menu-link-sublist-icon: url('data:image/svg+xml;utf8,<svg width="100%" height="100%" viewBox="0 0 16 16" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" xml:space="preserve" xmlns:serif="http://www.serif.com/" style="fill-rule:evenodd;clip-rule:evenodd;stroke-linejoin:round;stroke-miterlimit:2;"><g transform="matrix(0.88,0,0,0.88,1.3,1.3)"><path d="M11.505,8.838L11.071,9.271L8,6.2L4.929,9.271L4.496,8.837L8,5.333L11.505,8.838Z" style="fill:rgb(9,10,58);"/></g></svg>');
+
+  /* Increase navbar height. We set the variable here because it is used in multiple
        places in the layout, not just in the Navbar component we override. */
-    --ifm-navbar-height: 70px;
+  --ifm-navbar-height: 70px;
 
-    /* If we upgrade docusaurus from 2.0.0-beta.3, and yellow blockquotes introduced
+  /* If we upgrade docusaurus from 2.0.0-beta.3, and yellow blockquotes introduced
```
mfridman, line 53: Ah nice, they ended up fixing it. This is probably no longer relevant, for context:

https://github.com/facebook/docusaurus/issues/5244

mfridman (approved):
I have found admonitions useful in some rare cases, but we should apply them sparingly to avoid it being too distracting. 

I suggest we try them in a few places and see whether we like it before making sweeping changes.

lucperkins:
@mfridman I think I'm kind of in the opposite camp. If something is an "aside" I like it to stand out and for there to be visual cues. I do agree, though, that we should use the different categories judiciously. Another possibility might be to just have one admonition type that's styled in a way we really like.

mfridman:
> @mfridman I think I'm kind of in the opposite camp. If something is an "aside" I like it to stand out and for there to be visual cues. I do agree, though, that we should use the different categories judiciously. Another possibility might be to just have one admonition type that's styled in a way we really like.

Ye, I can't recall the example, but some docs I was on had an admonition every second paragraph and it was super distracting. I think we're all on the same page to use them appropriately and there is some value.

Also, I like the idea of having a single admonition that fits with our docs well.

lucperkins:
I've updated the blockquote aesthetic as well for the existing ones:

Screen Shot 2022-01-26 at 10 27 32 AM

PR #43, opened 2022-01-26
Add note on VS Code highlighting

lucperkins:


docs/editor-integration.mdx
```diff
@@ -33,12 +33,12 @@ problems, be sure to run a module-wide `buf lint` as part of your [CI](ci-cd/set
 
 ## Visual Studio Code  [![Visual Studio Marketplace Downloads](https://img.shields.io/visual-studio-marketplace/d/bufbuild.vscode-buf?color=2048ff&label=Buf&logo=visual-studio-code&style=flat-squar)](https://marketplace.visualstudio.com/items?itemName=bufbuild.vscode-buf)
 
-The Visual Studio Code extension can be downloaded from the in-editor extension browser under the name "Buf",
-or manually via [the extension page](https://marketplace.visualstudio.com/items?itemName=bufbuild.vscode-buf).
+The Visual Studio Code extension can be downloaded from the in-editor extension browser under the name "Buf"
+or manually via [the extension page](https://marketplace.visualstudio.com/items?itemName=bufbuild.vscode-buf). You need to have `buf` [installed](installation.mdx) to use it.
 
-To use VSCode integration, `buf` must be [installed](installation.mdx).
+Our Buf extension currently supports linting. As it doesn't support syntax highlighting, we recommend installing one of the [other available extensions][vs_code_protobuf_extensions] for that functionality.
```
johanbrandhorst, line 39: We've generally been very careful with our explicit recommendations that are related to other tools, especially when we plan on adding this functionality in the future. I wonder if we can just say something along the lines of "the roadmap includes full LSP support, formatting, syntax highlighting" etc which implies it's not there today without recommending anything else.

lucperkins: Okay, agreed that making this implicit rather than explicit is the way to go. Will change.

lucperkins: Done

johanbrandhorst (approved):

PR #47, opened 2022-01-27
Use Vale linter for spellchecking

lucperkins:
This PR adds spellchecking to the repo courtesy of the Vale linter and corrects a number of misspellings uncovered by the initial run.


mfridman (approved):

lucperkins:
Yeah, it can be swapped pretty easily, but Vale is by far the most powerful, as it can potentially encompass other tools like write-good and proselint, so it seems like the natural choice.

PR #51, opened 2022-01-29
Add i.e. and e.g. rules and fix instances

lucperkins:
This PR adds lint rules for i.e. and e.g. and replaces all current instances with other constructs (or removes them in some cases). I'm not 100% sure on all of the replacements and am open to suggestion. We're not using them _incorrectly_ per se but they're notoriously difficult for non-native speakers (and even native speakers) and there's pretty much never a time when they can't be replaced with something better.


docs/generate/managed-mode.md
```diff
@@ -4,7 +4,7 @@ title: Managed Mode
 ---
 
 **Managed Mode** is a [`buf.gen.yaml`](../configuration/v1/buf-gen-yaml.md) configuration option that tells `buf` to set all of the file
-options in your module according to an opinionated set of values suitable for each of the supported Protobuf languages (e.g. Go, Java, C#, etc.).
+options in your module according to an opinionated set of values suitable for each of the supported Protobuf languages, such as Go, Java, and C#.
```
mfridman, line 7: This sounds limiting to those languages, suggestion:

.. such as Go, Java, C# and others.

docs/tour/detect-breaking-changes.md
```diff
@@ -30,7 +30,7 @@ breaking:
 
 Next, you'll need to introduce a breaking change. First, make a change that's breaking at the
 [`WIRE`][wire] level. This is the most fundamental type of breaking change as it changes how the
-Protobuf messages are encoded in transit (i.e. "on the wire"). This type of breaking change affects
+Protobuf messages are encoded in transit and "on the wire." This type of breaking change affects
```
mfridman, line 33: Maybe drop the "and" here? The "in transit" equals "on the wire", I think i.e., loses its meaning on this one in particular.

.. Protobuf messages are encoded in transit ("on the wire").

docs/tour/generate-code.md
```diff
@@ -107,7 +107,7 @@ start/
 
 [Managed Mode](../generate/managed-mode.md) is a `buf.gen.yaml` configuration option that tells `buf`
 to set all of the file options in your module according to an opinionated set of values suitable for each of the
-supported Protobuf languages (e.g. Go, Java, C#, etc.). The file options are written *on the fly* by
+supported Protobuf languages, such as Go, Java, and C#. The file options are written *on the fly* by
```
mfridman, line 110: Similar comment as above.

docs/tour/use-managed-mode.md
```diff
@@ -10,7 +10,7 @@ our `.proto` files altogether.
 As discussed in [Generate Code](generate-code.md), **Managed Mode** is a
 [`buf.gen.yaml`](../configuration/v1/buf-gen-yaml.md) configuration option that tells `buf` to
 set all of the file options in your module according to an opinionated set of values suitable for each
-of the supported Protobuf languages (e.g. Go, Java, C#, etc.). The file options are written *on the fly*
+of the supported Protobuf languages, such as Go, Java, and C#. The file options are written *on the fly*
```
mfridman, line 13: Similar comment as above.

mfridman (approved):
Minor nits.

Anytime we can make our docs more accessible is a big win.

docs/generate/managed-mode.md
lucperkins: I disagree. "Such as" is equivalent to "for example." I don't think it would be read as an exclusive list.

docs/tour/detect-breaking-changes.md
lucperkins: Good call

PR #52, opened 2022-01-31
Update plugin references

timostamm:
We should be pointing to the plugin https://buf.build/grpc/plugins/go instead of https://buf.build/library/plugins/go-grpc,  
and to the template https://buf.build/grpc/templates/go instead of https://buf.build/library/template/go-grpc.


johanbrandhorst (approved):
🙏🏻

PR #58, opened 2022-02-01
[WIP] bazel docs

srikrsna-buf:
Related to: https://github.com/bufbuild/rules_buf/issues/6

Adds **Build Systems** section right after **Editor Integration**.  Contains documentation for [rules_buf](https://github.com/bufbuild/rules_buf).


docs/build-systems/bazel/setup.md
```diff
@@ -0,0 +1,132 @@
...
+    targets = [":foo_proto"],
+    config = ":buf.yaml",
+)
+```
+
+This can be run as:
+```
+$ bazel test :foo_proto_breaking
+```
+
```
srikrsna-buf, line 107: @amckinney 

Should we include the full list of [attributes](https://github.com/bufbuild/rules_buf/tree/gazelle/buf#buf_breaking_test) here? Since they are generated I'm tempted to include the link.

I hate it when a doc site takes me to github and vice versa.

docs/build-systems/bazel/gazelle.md
```diff
@@ -0,0 +1,121 @@
...
+Run the following command to list the generated breaking rules:
+```terminal
+$ bazel query 'kind(buf_breaking_test, //...)'
+```
+
+This mimics running `buf breaking` on a module. This is the most accurate way to check for breaking changes. However, depending on multiple targets at once is an anti-pattern in bazel. To do this the bazel way we have package mode.
+
+### Package mode
+
+Package mode generates `buf_breaking_test` rule for each of the `proto_library` rule. Allowing one to test only the `proto_library` that has changed. This could be useful in large codebases where checking the module can be time consuming.
```
amckinney, line 82: > This could be useful in large codebases where checking the module can be time consuming.

I don't know if this is an accurate reason - breaking change detection is _fast_, even for large modules like GoogleAPIs. We might want to remove this note entirely.

docs/build-systems/bazel/gazelle.md
```diff
@@ -0,0 +1,121 @@
...
+        ├── bar.proto
+        └── BUILD
+```
+#### Module mode
+
+Single `buf_breaking_test` rule is generated in `BUILD`. If a breaking change occurs in either `foo.proto` or `bar.proto` the test will catch that. Even if `foo.proto` is deleted the test will catch that.
+
+#### Package mode
+
+`buf_breaking_test` rules are generated in both `foo/v1/BUILD` and `bar/v1/BUILD` against their respective `proto_library` targets. If a breaking change occurs in either `foo.proto` or `bar.proto` the tests will catch that. But if either `foo.proto` or `bar.proto` is deleted the tests will fail to catch that.
```
amckinney, line 121:
```suggestion
`buf_breaking_test` rules are generated in both `foo/v1/BUILD` and `bar/v1/BUILD` against their respective `proto_library` targets. If a breaking change occurs in either `foo.proto` or `bar.proto` this tests will detect it. However, if either `foo.proto` or `bar.proto` is deleted the tests will fail to detect it.
```

amckinney, line 121: > However, if either `foo.proto` or `bar.proto` is deleted the tests will fail to detect it.

Can we add more explanation here? Why is that the case / why is package mode unable to do this?

amckinney (commented):
This is awesome! Well done so far. A bunch of structural things and rephrasing, but this is on the right track.

docs/build-systems/bazel/gazelle.md
```diff
@@ -0,0 +1,121 @@
...
+)
+```
+
+Export the `buf.yaml` file by adding `exports_files(["buf.yaml"])` to the `BUILD` file.
+
+> For workspaces do this for each `buf.yaml` file.
+
+Now run Gazelle
+```terminal
+$ bazel run //:gazelle
```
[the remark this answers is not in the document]
srikrsna-buf: The readme says terminal removes `$ ` if copied. We use the `$` in other places: https://github.com/bufbuild/docs.buf.build/blob/6a2ca4bb6ea4610551da3bd19aec380dbd2b2c13/docs/ci-cd/setup.md#L139

docs/build-systems/bazel/setup.md
```diff
@@ -0,0 +1,132 @@
...
+## Setup
+
+Add the following snippet to the `WORKSPACE` file replacing the `<SHA256>` and `<VERSION>` with those of a [specific release](https://github.com/bufbuild/rules_buf/releases):
+```starlark title="WORKSPACE"
+load("@bazel_tools//tools/build_defs/repo:http.bzl", "http_archive")
+
+http_archive(
+    name = "rules_buf",
+    sha256 = "<SHA256>",
```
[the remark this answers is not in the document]
srikrsna-buf: Okay once we tag a version we will be able to do that. But we may have to update it at least until buf v1 is launched.

docs/build-systems/bazel/setup.md
```diff
@@ -0,0 +1,132 @@
...
+    urls = [        
+        "https://github.com/bufbuild/rules_buf/releases/download/<VERSION>/rules_go-<VERSION>.zip",
+    ],
+)
+
+load("@rules_buf//buf:repositories.bzl", "rules_buf_dependencies", "rules_buf_toolchains")
+
+rules_buf_dependencies()
+
+rules_buf_toolchains()
```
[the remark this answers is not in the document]
srikrsna-buf: `rules_go` uses new lines and `rules_proto` doesn't 😞

docs/build-systems/bazel/overview.md
```diff
@@ -0,0 +1,191 @@
...
+http_file(
+      name = "buf_module",
+      urls = ["http://s3-us-east-1.amazonaws.com/bucket/foo/bar/images/${COMMIT}/image.bin"],
+      sha256 = "...",
+)
+``` 
+
+This file can be referenced from `buf_breaking_test`. The commit and sha256 need to be updated as needed.
+
+> For repositories using `buf.work.yaml` that reference multiple `buf.yaml` files. A single image file should be maintained for each `buf.yaml` file. This is true for both module and package level granularity of `buf_breaking_test`.
```
amckinney, line 152: This is great, thanks for adding this. Recommending `testdata` seems fine - I'm not sure where else it makes sense to deposit an `image.bin` only used for testing.

amckinney (approved):
Very nice 👏  Let's get @lucperkins' eyes on this too. I assume we won't land this until we've made the repository public and ready for use.

lucperkins:
@srikrsna-buf Awesome. Really glad to be getting the ball rolling on this. For now, I'm going to undo the changes to `sidebar.json` and put the new docs into a `drafts` directory. Basically I need to go over the wording and make some very tech-writer-y changes and that's typically more efficiently done in a separate PR. But the substance here is fantastic!

PR #59, opened 2022-02-01
Minor tweaks and edits

robbertvanginkel:
Some minor tweaks based on my pass over the docs.

### DIFF
--- a/docs/build/usage.md
+++ b/docs/build/usage.md
@@ -168,7 +168,7 @@ the authors are being honest, we can't remember if it's the first `-I` or second
 we have outlawed this in our own builds for a long time.
 
 While the above example is relatively contrived, the common error that comes up is when you
-have vendored `.proto` files. For example, [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway/tree/master/third_party/googleapis/google)
+have vendored `.proto` files. For example, [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway/tree/cc01a282127b54a81f92d6b8e8fb8971dab8be9b/third_party/googleapis)
 has it's own copy of the [google.api](https://github.com/googleapis/googleapis/tree/master/google/api) definitions it needs.
 While these are usually in sync, the `google.api` schema can change. If we allowed the following:
 
@@ -223,7 +223,7 @@ However, `buf build` also supports outputting [FileDescriptorSets](https://githu
 and [Images](../reference/images.md), which is Buf's custom extension of the FileDescriptorSet. Better yet, these outputs
 can be formatted in a variety of ways.
 
-Per the [input documentation](../reference/inputs.md), `buf build` can deduce the output format by the file extension. For example,
+Per the [input documentation](../reference/inputs.md#automatically-derived-formats), `buf build` can deduce the output format by the file extension. For example,
 
 ```sh
 $ buf build -o image.bin
--- a/docs/build/usage.md
+++ b/docs/build/usage.md
@@ -223,7 +223,7 @@ However, `buf build` also supports outputting [FileDescriptorSets](https://githu
 and [Images](../reference/images.md), which is Buf's custom extension of the FileDescriptorSet. Better yet, these outputs
 can be formatted in a variety of ways.
 
-`buf build` can deduce the output format by the file extension. For example,
+`buf build` can deduce the output format by the file extension, see the documentation on [automatically derived formats](../reference/inputs.md#automatically-derived-formats). For example,
 
 ```sh
 $ buf build -o image.bin
--- a/docs/build/usage.md
+++ b/docs/build/usage.md
@@ -223,7 +223,7 @@ However, `buf build` also supports outputting [FileDescriptorSets](https://githu
 and [Images](../reference/images.md), which is Buf's custom extension of the FileDescriptorSet. Better yet, these outputs
 can be formatted in a variety of ways.
 
-Per the [input documentation](../reference/inputs.md#automatically-derived-formats), `buf build` can deduce the output format by the file extension. For example,
+`buf build` can deduce the output format by the file extension. For example,
 
 ```sh
 $ buf build -o image.bin
--- a/docs/ci-cd/setup.md
+++ b/docs/ci-cd/setup.md
@@ -47,13 +47,13 @@ import TabItem from '@theme/TabItem';
 PROJECT=<your-project-name>
 # Use your desired buf version
 BUF_VERSION=1.0.0-rc12
-# buf will be cached to ~/.cache/your-project-name.
-CACHE_BIN=$HOME/.cache/$(PROJECT)
+# buf will be installed to ~/bin/your-project-name.
+BIN_DIR=$HOME/bin/$(PROJECT)
 
 curl -sSL \
 	"https://github.com/bufbuild/buf/releases/download/v$BUF_VERSION/buf-$(shell uname -s)-$(shell uname -m)" \
-	-o "$CACHE_BIN/buf"
-chmod +x "$CACHE_BIN/buf"
+	-o "$BIN_DIR/buf"
+chmod +x "$BIN_DIR/buf"
 ```
 
 This script sends a request to the `buf` GitHub Releases using [`curl`](https://curl.se/docs)
--- a/docs/configuration/v1/buf-gen-yaml.md
+++ b/docs/configuration/v1/buf-gen-yaml.md
@@ -172,6 +172,9 @@ managed:
       - buf.build/googleapis/googleapis
     override:
       buf.build/acme/weather: github.com/acme/weather/gen/proto/go
+  override:
+    JAVA_PACKAGE:
+      acme/weather/v1/weather.proto: "org"
 plugins:
   - name: go
     out: gen/proto/go
--- a/docs/generate/usage.md
+++ b/docs/generate/usage.md
@@ -74,8 +74,8 @@ $ buf generate
 ```
 
 You can also run `buf generate` on a module by specifying the filepath to the
-directory containing the `buf.yaml`. For example, if you have a `buf.yaml` in
-directory foo, you can generate using:
+directory containing the proto root. For example if your protos are in directory
+`foo`:
 
 ```sh
 $ buf generate foo
--- a/docs/generate/usage.md
+++ b/docs/generate/usage.md
@@ -73,9 +73,9 @@ To generate for the input in your current directory, simply run:
 $ buf generate
 ```
 
-You can also run `buf generate` on a module by specifying the filepath to the
-directory containing the proto root. For example if your protos are in directory
-`foo`:
+You can also run `buf generate` on an input by specifying the filepath to the
+directory containing the root of your `.proto` definitions. For example if all of
+your `.proto` files are in directory `foo`:
 
 ```sh
 $ buf generate foo
--- a/docs/generate/usage.md
+++ b/docs/generate/usage.md
@@ -67,8 +67,18 @@ invocation, with results from each invocation combined before writing the result
 
 ## Run generate
 
+To generate for a buf module in your current directory, simply run:
+
 ```sh
 $ buf generate
+```
+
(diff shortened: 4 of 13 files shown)


docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -172,6 +172,9 @@ managed:
       - buf.build/googleapis/googleapis
     override:
       buf.build/acme/weather: github.com/acme/weather/gen/proto/go
+  override:
+    JAVA_PACKAGE:
+      acme/weather/v1/weather.proto: "org"
```
robbertvanginkel, line 177: This is because below on line 270 
> `override`
> This is a list of per-file overrides for each modifier. In the example provided above, an override for `acme/weather/v1/weather.proto` is set for the `java_package_prefix`
modifier to be `org` instead of `com`. This will set `org` as the package prefix for **only** the specific `acme/weather/v1/weather.proto` file and **not** for the rest of the module.

Otherwise the "example provided above" refers to nothing. The example is taken from `docs/generate/managed-mode.md`.

docs/generate/usage.md
```diff
@@ -67,10 +67,6 @@ invocation, with results from each invocation combined before writing the result
 
 ## Run generate
 
-You can run `buf generate` on your module by specifying the filepath to
-the directory containing the `buf.yaml`. In the above example, you can target
-the `buf.build/acme/petapis` input defined in the current directory like so:
```
robbertvanginkel, line 72: There is no reference to petapis in this file or example. I wasn't really sure what to do about 

> module by specifying the filepath to the directory containing the `buf.yaml`.

because as far as I understand generate is always run in the directory with the yaml, unless using workspaces.

docs/reference/images.md
```diff
@@ -6,7 +6,7 @@ title: Images
 Throughout the documentation, you will occasionally see references to **Images**. We'll go over what
 Images are, how they are used, and the various options associated with them here.
 
-## Protobuf plugins: how they work
```
robbertvanginkel, line 9: May not want to change titles because of anchors, but this read a bit odd. Happy to discard, same for other change in this file.

docs/tour/add-a-dependency.md
```diff
@@ -125,7 +125,7 @@ you can specify it like so:
  version: v1
  deps:
 -  - buf.build/googleapis/googleapis
-+  - buf.build/googleapis/googleapis:1c473ad9220a49bca9320f4cc690eba5
```
robbertvanginkel, line 128: This comment doesn't exist on `googleapis/googleapis`, it did on `beta/googleapis`: https://buf.build/beta/googleapis/history/1c473ad9220a49bca9320f4cc690eba5

This changes it to the commit for the same tag on `googleapis/googleapis`.

docs/tour/view-generated-documentation.md
```diff
@@ -62,8 +62,8 @@ the `buf.md`.
 ## 8.2 Package Documentation {#package-documentation}
 
 As you can see from the module documentation page, both the `pet.v1` and `google.type`
-packages are available as links. Click on the `pet.v1` link to navigate to its package
-documentation at [https://buf.build/acme/petapis/docs/main/pet.v1](https://buf.build/acme/petapis/docs/main/pet.v1).
```
robbertvanginkel, line 66: Since `main` here refers to the end state of the tutorial, there is no `google.type` package in it. Pointing to an older commit (that is the result from buf push on the line above) rather than the main branch seemed reasonable.

johanbrandhorst (approved):
Really nice changes IMO, but wait for Luc to review

docs/configuration/v1/buf-gen-yaml.md
robbertvanginkel: @amckinney separte note I had here on the same in `docs/generate/managed-mode`

Here we have `managed.go_package_prefix.override` but also `managed.override.JAVA_PACKAGE` and `managed.java_package_prefix` ? This is probably how our config is setup, but having overrides in different places for different languages seems odd at first sight. The example also contains `JAVA_PACKAGE` but the description for `managed.override` is:

> This is a list of per-file overrides for each modifier. In the example provided above, an override for acme/weather/v1/weather.proto is set for the java_package_prefix modifier to be org instead of com. This will set org as the package prefix for only the specific acme/weather/v1/weather.proto file and not for the rest of the module.

How does `JAVA_PACKAGE` here relate to `java_package_prefix`?

docs/build/usage.md
```diff
@@ -168,7 +168,7 @@ the authors are being honest, we can't remember if it's the first `-I` or second
 we have outlawed this in our own builds for a long time.
 
 While the above example is relatively contrived, the common error that comes up is when you
-have vendored `.proto` files. For example, [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway/tree/master/third_party/googleapis/google)
+have vendored `.proto` files. For example, [grpc-gateway](https://github.com/grpc-ecosystem/grpc-gateway/tree/cc01a282127b54a81f92d6b8e8fb8971dab8be9b/third_party/googleapis)
```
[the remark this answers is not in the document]
amckinney: Works for me. It'd be great if we could find another popular repository that vendors Protobuf though. That way we clearly demonstrate how much of an improvement this is.

docs/build/usage.md
```diff
@@ -223,7 +223,7 @@ However, `buf build` also supports outputting [FileDescriptorSets](https://githu
 and [Images](../reference/images.md), which is Buf's custom extension of the FileDescriptorSet. Better yet, these outputs
 can be formatted in a variety of ways.
 
-Per the [input documentation](../reference/inputs.md), `buf build` can deduce the output format by the file extension. For example,
+Per the [input documentation](../reference/inputs.md#automatically-derived-formats), `buf build` can deduce the output format by the file extension. For example,
```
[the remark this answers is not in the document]
amckinney: We could invert this a bit and make it more direct.
```suggestion
`buf build` can deduce the output format by the file extension. For example,
```

We can move the `"For more details, see the [input documentation](...)` immediately after the `jq` example and before `ImageExtension` is called out (or somewhere else entirely).

docs/ci-cd/setup.md
```diff
@@ -47,13 +47,13 @@ import TabItem from '@theme/TabItem';
 PROJECT=<your-project-name>
 # Use your desired buf version
 BUF_VERSION=1.0.0-rc12
-# buf will be cached to ~/.cache/your-project-name.
-CACHE_BIN=$HOME/.cache/$(PROJECT)
+# buf will be installed to ~/bin/your-project-name.
```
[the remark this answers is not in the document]
amckinney: I know this name was intentionally used so that it was symmetrical with [buf-example](https://github.com/bufbuild/buf-example/blob/b4ad2ea50b2c579a32f52b36561db78813212e42/Makefile#L36).

I agree that it's bit of a misnomer in this case, but I don't know if this makes it any more clear. I guess I could get behind `bin` and `BIN_DIR` though because that's actually what it's doing.

docs/configuration/v1/buf-gen-yaml.md
amckinney: Yup, we didn't update this document to include the `override` key in general - good catch.

The difference is that the `top-level` override key applies to individual files _explicitly_ with the value, whereas the `{go,java}_package_prefix` keys apply a package _prefix_ to specific modules, but continue to format the `{go,java}_package` values according to the Managed Mode opinions.

For reference, see the implementation for [go](https://github.com/bufbuild/buf/blob/9c0d9e9c7684c2a05db336b1215ac61dfc10d454/private/bufpkg/bufimage/bufimagemodify/go_package.go#L69) and [java](https://github.com/bufbuild/buf/blob/9c0d9e9c7684c2a05db336b1215ac61dfc10d454/private/bufpkg/bufimage/bufimagemodify/java_package.go#L148).

docs/generate/usage.md
amckinney: We refer to `buf.build/acme/weather` in plenty of other files though. It's simply included to be an example for a module that can be used as input.

I agree that this information doesn't need to be here, but we should add the module reference (with `buf.build/acme/petapis`) to the `Common use cases` below.

> because as far as I understand generate is always run in the directory with the yaml, unless using workspaces.

You can run `buf generate` on an input that doesn't define a `buf.yaml`. Just like `buf build`, `buf` will assume that the directory is a source that defines `.proto` sources relative to that root.

docs/reference/images.md
amckinney: Breaking compatibility for the anchor isn't the end of the world if this one isn't linked directly from other places in the docs. I doubt anyone has this sub-section pinned, for example.

We might want to un-capitalize  `Plugins` though so it's consistent with the other headers.

lucperkins:
Good catches all around 👍🏼

lucperkins (approved):

docs/configuration/v1/buf-gen-yaml.md
robbertvanginkel: The override key is included though, see https://docs.buf.build/configuration/v1/buf-gen-yaml#override-1. Its a little hard to follow because `managed.go_package_prefix.override` and `managed.override` are just below each other and the subheadings don't make this very clear:
Screen Shot 2022-02-02 at 11 13 25 AM

I think understand where `JAVA_PACKAGE` comes from now though 👍

docs/reference/images.md
```diff
@@ -6,7 +6,7 @@ title: Images
...
+## How Protobuf Plugins work
```
robbertvanginkel, line 9:
```suggestion
## How Protobuf plugins work
```

docs/generate/usage.md
robbertvanginkel: I think I was confused as the two examples in text (subdir and module) are removed from the example, which specifies nothing which defaults to the working directory. Updated and also added to the "common use cases" below, lmk if this makes sense.

docs/configuration/v1/buf-gen-yaml.md
amckinney: Yea I agree that those sub-headers need to be made more clear. This is a problem elsewhere in the docs, so we can leave it for a follow-up. Let's create an issue for it if one doesn't already exist.

docs/generate/usage.md
```diff
@@ -67,14 +67,20 @@ invocation, with results from each invocation combined before writing the result
...
+To generate for a buf module in your current directory, simply run:
```
amckinney, line 70: Use the word `input` so that it's clear it doesn't have to be a module (like elsewhere in the docs):
```suggestion
To generate for the input in your current directory, simply run:
```

docs/generate/usage.md
```diff
@@ -67,14 +67,20 @@ invocation, with results from each invocation combined before writing the result
...
 
 ```sh
 $ buf generate
 ```
 
+You can also run `buf generate` on a module by specifying the filepath to the
+directory containing the proto root. For example if your protos are in directory
+`foo`:
```
amckinney, line 78: I'd make this a bit more explicit.

```suggestion
You can also run `buf generate` on an input by specifying the filepath to the
directory containing the root of your `.proto` definitions. For example if all of
your `.proto` files are in directory `foo`:
```

docs/generate/usage.md
```diff
@@ -105,6 +111,9 @@ The following section describes several common cases for `buf generate`:
 # Uses the current directory as input, and assumes a `buf.gen.yaml` also exists in the current directory.
 $ buf generate
 
+# Uses the buf.build/acme/petapis module as input, using a `buf.gen.yaml` in the current directory.
+$ buf generate buf.build/acme/petapis
```
amckinney, line 115: Nice 👏

amckinney (approved):
Nice edits, thanks for shepherding this across!

docs/configuration/v1/buf-gen-yaml.md
amckinney: Relevant issue here: https://github.com/bufbuild/docs.buf.build/issues/64

PR #63, opened 2022-02-02
Document the track concept in BSR

lucperkins:
We'll certainly need more practical docs for using tracks and to improve the wording here, but I wanted to achieve some consensus first.

**Note**: this would be pushed only in conjunction with the tracks release.

Deploy preview: https://deploy-preview-63--buf.netlify.app/bsr/overview#referencing-a-module


docs/bsr/overview.md
```diff
@@ -106,25 +106,77 @@ The `buf` CLI automatically resolves the module(s) specified in the `deps` list.
...
+    {separator: "/"},
+    {label: "repository", kind: "variable"},
+    {separator: ":"},
+    {label: "commit|tag|track", kind: "variable"},
+  ]
+} />
+
+### Commit
+
+Every push of new content to a BSR repository is associated with a **commit** that identifies that
```
tomchwojkofrank, line 133: In the `Tag` section, you mention that our `tag` is similar to, but not the same as a Git tag. Do we want to use the same kind of phrasing here for `commit`?

docs/bsr/overview.md
```diff
@@ -106,25 +106,77 @@ The `buf` CLI automatically resolves the module(s) specified in the `deps` list.
...
+`9f6b71c31322443f8c08b9fa3c7261ae` commit, these two references refer to the same module snapshot:
+
+* `buf.build/acme/petapis:v0.1.0`
+* `buf.build/acme/petapis:9f6b71c31322443f8c08b9fa3c7261ae`
+
+### Track
+
+A **track** is a set of [commits](#commit) that share a name that you provide. You might create
+`development` and `production` tracks for a module, for example, to group commits based on the
+intended environment.
```
WillAbides, line 158: It might help to also give the example of creating a track for each branch in your git repository.

lucperkins: Good call. Done.

docs/bsr/overview.md
```diff
@@ -106,25 +106,77 @@ The `buf` CLI automatically resolves the module(s) specified in the `deps` list.
...
+
+Tracks are somewhat like [branches] in version control systems. As with branches, commits in tracks
+are in reverse chronological order. But unlike branches, commits in a BSR track have no logical
+connection to one another; you can assign any commits you like to any track.
```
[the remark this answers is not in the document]
lucperkins: Good catch. Quite an important distinction. Fixed.

docs/bsr/overview.md
lucperkins: Fair point. In this case, I think that the explanation below makes this sufficiently clear, but we can say more if users get tripped up.

PR #68, opened 2022-02-04
Delete buf protoc documentation

bufdev:
See https://github.com/bufbuild/buf/pull/915.


doriable (approved):

doriable:
Would this be merged after the release without `buf protoc`, or should we preemptively remove this? Personally, I think it's fine to merge this now since we are clearly making plans to remove `buf protoc` and I'm in favour of not pointing folks to this anymore.

bufdev:
I think we merge now, no need to keep documenting.

PR #71, opened 2022-02-05
Standardize on sh highlighting

lucperkins:
Just a small ergonomic thing. The minor benefit of using `sh` is that you get syntax highlighting in Markdown in most editors. Just saw multiple shells being used.

@mfridman If there was a specific reason for choosing `terminal`, let me know and I'm happy to update/revert.

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,20 +7,20 @@ https://docs.buf.build.
 
 To get started, make sure you have [Node] installed:
 
-```terminal
+```sh
 brew install node
 make install
 ```
 
 To run a local dev server, run:
 
-```terminal
+```sh
 make
 ```
 
 or
 
-```terminal
+```sh
 make run
 ```
 
@@ -30,7 +30,7 @@ having to restart the server.
 
 To view the fully built product, you can run:
 
-```terminal
+```sh
 make serve
 ```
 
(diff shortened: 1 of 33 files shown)


srikrsna-buf:
If I have to choose between syntax highlighting vs the copy stripping `$ `, I will choose the later. If we can get both that would be great.

mfridman:
@lucperkins We chose terminal as an identifier to strip prefixes when copying commands. See this function here:

https://github.com/bufbuild/docs.buf.build/blob/f7edfbd095a55b70465183605d554c6037f45507/src/theme/CodeBlock/index.tsx#L50-L67

The idea was .. use `terminal` if you want to strip the prefix and `sh` / `bash` otherwise. I thought we also applied the bash/sh syntax highlighting but need to double-check the .css file.

mfridman:
Oh, and more importantly it trimmed the output as well, so when you hit copy it'll only copy the command portion.

CleanShot 2022-02-05 at 00 47 43@2x

lucperkins:
@mfridman Oh duh, that's right, my bad 🙄 Completely slipped my mind. I'll go ahead and rethink these changes, then.

PR #80, opened 2022-02-08
Remove fields that are no longer in the `buf.lock`

doriable:
These fields have since been removed from the `buf.lock`.


amckinney (approved):
I wonder if we should include a deprecation note on the page that describes which fields were removed. Once we release `v1` and (most) of our users move onto it, the note won't be relevant anymore but it might help people between now and then.

Thinking out loud (cc @lucperkins)

doriable:
Hmmm, that's a good point. On one hand, I wonder if it'll be more confusing to users, since we want our "gold standard" to be `v1`. It's also tough in this case, because we don't have versioned docs, so someone on an older CLI version might find the lack of fields confusing... I added a note at the bottom that might help but am open to thoughts on this :)

docs/configuration/v1/buf-lock.md
```diff
@@ -60,10 +57,11 @@ deps:
...
     repository: paymentapis
-    branch: main
     commit: 5173e5cfeb904508839378050d95e1de
-    digest: b1-tteutIW3X3VYL65FfLiqkmW_DIDUuqRbyuxOSmscB5Q=
-    create_time: 2021-06-03T17:53:54.361543Z
 ```
 
+Note that older versions of `buf` may include `branch`, `commit`, `digest`, and `create_time`
+as a part of the dependencies. Older `buf.lock` files are updated without the fields when
+`buf mod update` is run on a newer version of `buf`.
```
amckinney, line 65: Let's include this as a block in the note format we use elsewhere.
```suggestion
> Older versions of `buf` may include `branch`, `commit`, `digest`, and `create_time`
> as a part of the dependencies. The latest `buf.lock` file will not include these fields when
> `buf mod update` is run on a newer version of `buf`.
```

We can land this as-is, but flagging @lucperkins for potential language updates in a follow up (if ever).

lucperkins:
@amckinney Good question. Part of me thinks warnings are always good, part of me wonders thinks lockfiles are meant to be opaque. Do we anticipate user confusion?

amckinney:
That's a good point, too. `buf.lock` files are definitely meant to be opaque. I assume _some_ users will be surprised by it, but maybe they don't care. We could start without it, and only include it if we get enough questions about it.

I'm fine with either way here - I don't see much of a harm either way.

docs/configuration/v1/buf-lock.md
```diff
@@ -60,10 +57,11 @@ deps:
...
     owner: acme
     repository: paymentapis
-    branch: main
     commit: 5173e5cfeb904508839378050d95e1de
-    digest: b1-tteutIW3X3VYL65FfLiqkmW_DIDUuqRbyuxOSmscB5Q=
-    create_time: 2021-06-03T17:53:54.361543Z
 ```
 
+> Older versions of `buf` may include `branch`, `commit`, `digest`, and `create_time`
+> as a part of the dependencies. The latest `buf.lock` file will not include these fields when
```
lucperkins, line 64: I'd rephrase this second sentence:

"Your `buf.lock` shouldn't include these fields if you've run `buf mod update` with a newer version of `buf`."

lucperkins:
@amckinney Okay, I think including is fine. One small wording nit and merge at will.

lucperkins (approved):

PR #81, opened 2022-02-08
Restructure managed mode docs

### LINKED ISSUE #96: Restructure Managed Mode doc
The [Managed Mode](https://docs.buf.build/generate/managed-mode) doc is a useful reference, but we should restructure it to make it more guide like for users of specific languages. That is, the target "persona" for that doc should be a developer in a specific language (Go, Java, etc.).

lucperkins:
Affected page: https://deploy-preview-81--buf.netlify.app/generate/managed-mode.

Fixes #96

This is a pretty drastic rewrite so feel free to be critical. I think there are some nuances about MM that I'm still missing but I'm hoping that the language-oriented shift better matches the interests of users.


docs/generate/managed-mode.md
```diff
@@ -3,9 +3,18 @@ id: managed-mode
...
-**Managed Mode** is a [`buf.gen.yaml`](../configuration/v1/buf-gen-yaml.md) configuration option that tells `buf` to set all of the file
-options in your module according to an opinionated set of values suitable for each of the supported Protobuf languages, such as Go, Java, and C#.
-The file options are written *on the fly* so that they never have to be written in the Protobuf source file itself.
+**Managed Mode** is a [`buf.gen.yaml`](../configuration/v1/buf-gen-yaml.md) configuration option
+that tells the `buf` CLI to set the file options in your module according to an opinionated set of
+values. Buf has chosen such values for these languages:
+
+* [C++](#cc_enable_arenas)
+* [Java](#java_multiple_files)
+* [Go](#go_package_prefix)
```
johanbrandhorst, line 12: I still don't think this is complete, we set options for various other languages too (PHP, C#), see further down on this page. What is it we're trying to express with this top level summary? CC @amckinney who is closest to this.

lucperkins: Yeah, Alex set me straight on this one. I created issue #96 for this, as I think we need a bolder intervention here.

docs/generate/managed-mode.md
```diff
@@ -3,133 +3,148 @@ id: managed-mode
...
+compiler in some circumstances, such as [`go_package`](#go_package_prefix) when generating Go code.
+These options have been a pain point in Protobuf development for many years because they require API producers to set
+values that [don't really belong](#background) in API definitions.
+
+You can avoid setting these file options when [generating] code from Protobuf sources by enabling
+**managed mode** in your [`buf.gen.yaml`](../configuration/v1/buf-gen-yaml.md) configuration file.
+When managed mode is [enabled](#enabled), the `buf` CLI sets specified file options on the fly during
+code generation so that you don't need to hard-code them in your `.proto` files.
+
+Managed mode provides options for these languages:
```
mfridman, line 17: I think this should also include swift?

docs/generate/managed-mode.md
```diff
@@ -3,133 +3,148 @@ id: managed-mode
...
+* [C++](#cpp)
+* [C#](#csharp)
+* [Go](#go)
+* [Java](#java)
+* [Objective-C](#objective-c)
+* [PHP](#php)
+* [Ruby](#ruby)
+
+> If you're generating code for a language that isn't on this list, managed mode has no implications
+> and enabling it is likely to have no effect.
```
mfridman, line 28: Should we use stronger language here? ".. and enabling it will have no effect."

docs/generate/managed-mode.md
```diff
@@ -207,30 +286,81 @@ option objc_class_prefix = "AWX";
...
 ```
 
-> Some options, such as `cc_enable_arenas` and `optimize_for`, are excluded from this list because managed mode agrees with the default
-> values specified by [google/protobuf/descriptor.go](https://github.com/protocolbuffers/protobuf/blob/b650ea44b10133008baaea7488360c5b95c93c7b/src/google/protobuf/descriptor.proto#L385).
-> If you disagree with the default values, you can override these option values, which is described in the [next section](#file-option-overrides).
+But with the `buf` CLI, you wouldn't ever _see_ this file. Those options would be written on the fly
+and used as part of the generation process.
 
-## File option overrides
+## Background
```
mfridman, line 296: We've historically placed "background" at the top, curious about this change? Not against it, because folks typically want docs to tell them what to do, and as a bonus they can get additional context .. but that additional context shouldn't get in the way of the docs. 

Just curious what your take on this is.

lucperkins: In general, I think we should strive to make a more clean separation between practical docs and reference. See also #112. In this particular case, MM is complex enough that having a reference-y section on the same page is indeed necessary, but the docs is generally meant to be practical and thus to enable users to accomplish specific goals (and this PR is basically meant to clarify those goals further by dividing things up by language). But I made sure to give users an "on-ramp" to the Background section in the initial paragraph because I think you're right that the % of users who will need that background is pretty high.

docs/generate/managed-mode.md
lucperkins: An excellent question! Alex M pointed me to [this issue](https://github.com/bufbuild/core/issues/1315) where the consensus was that Apple counsels against `swift_prefix` and so it's not supported in MM. I added a note near the top that makes that clear.

docs/generate/managed-mode.md
lucperkins: Yeah, we should. Done.

docs/generate/managed-mode.md
mfridman: Ye that's fair. Overall looks good.

I do hope we can keep adding and making it easy for users to find these background sections, either via links or referring to them later on in the page. I personally (and others) have found them extremely useful.

mfridman (approved):
lgtm.

PR #91, opened 2022-02-10
Add JS linting to docs site

lucperkins:
I don't think we need to add these in CI or anything like that just yet, but it'd be nice for us to have a lever that we can periodically pull over time to keep things neat and tidy. Ordinarily using a system like Docusaurus would make this sort of thing superfluous but since we're using custom components we should treat it like a full-fledged JS project. When attempting to upgrade Docusaurus in #72 I ran into lots of little TypeScript gotchas that I'd like to avoid in the future.

FYI most of these changes were made by `prettier`.


mfridman (approved):
I have zero opinion on how code should be formatted, as long as we're consistent then 👍

PR #118, opened 2022-03-01
Add documentation for build's `--type` flag

robbertvanginkel:
Added in https://github.com/bufbuild/buf/pull/949, released in https://github.com/bufbuild/buf/releases/tag/v1.1.0.

### DIFF
--- a/docs/build/usage.md
+++ b/docs/build/usage.md
@@ -285,12 +285,12 @@ $ buf build --path path/to/foo.proto --path path/to/bar.proto
 
 ## Limit to specific types
 
-The output FileDescritptorSet or image from a `buf build` will contain all types
-declared in the module. For advanced usecases, users may want an image or
-FileDescriptorSet containing only a subset of types. Introduced in `v1.1`, the
-`--type` flag accepts fully qualified protobuf names and limits the output image
+The output `FileDescriptorSet` or image from a `buf build` will contain all types
+declared in the module. For advanced use cases, users may want an image or
+FileDescriptorSet containing only a subset of types. Introduced in `v1.1.0`, the
+`--type` flag accepts fully qualified Protobuf names and limits the output image
 to contain only descriptors required to represent those types and their required
-dependencies. The type flag supports messages, enums and services.
+dependencies. The type flag supports messages, enums, and services.
 
 ```terminal
 $ buf build --type pkg.foo.Bar
@@ -315,7 +315,7 @@ is needed to accurately and completely represent that descriptor. These include:
       in which the message is defined
 
 As an example, consider the these sample protos:
-```proto
+```protobuf
 --- foo.proto ---
 package pkg;
 message Foo {
--- a/docs/build/usage.md
+++ b/docs/build/usage.md
@@ -283,6 +283,75 @@ command:
 $ buf build --path path/to/foo.proto --path path/to/bar.proto
 ```
 
+## Limit to specific types
+
+The output FileDescritptorSet or image from a `buf build` will contain all types
+declared in the module. For advanced usecases, users may want an image or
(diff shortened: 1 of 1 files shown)


docs/build/usage.md
```diff
@@ -283,6 +283,75 @@ command:
 $ buf build --path path/to/foo.proto --path path/to/bar.proto
 ```
 
+## Limit to specific types
+
+The output FileDescritptorSet or image from a `buf build` will contain all types
```
doriable, line 288: `%s/FileDescritptorSet/FileDescriptorSet`. As a side note, I think we show `FileDescriptorSet` with the code backticks elsewhere in the documentation, so we may want to do that here also.

docs/build/usage.md
```diff
@@ -283,6 +283,75 @@ command:
...
+declared in the module. For advanced usecases, users may want an image or
```
doriable, line 289: `%s/usecases/use cases`

docs/build/usage.md
```diff
@@ -283,6 +283,75 @@ command:
...
+FileDescriptorSet containing only a subset of types. Introduced in `v1.1`, the
```
doriable, line 290: `%s/v1.1/v1.1.0`? i think this could go either way, tbh.

docs/build/usage.md
```diff
@@ -283,6 +283,75 @@ command:
...
+`--type` flag accepts fully qualified protobuf names and limits the output image
```
doriable, line 291: I believe "Protobuf" is capitalised in other parts of our documentation.

docs/build/usage.md
```diff
@@ -283,6 +283,75 @@ command:
...
+to contain only descriptors required to represent those types and their required
+dependencies. The type flag supports messages, enums and services.
```
doriable, line 293: Not to be "that guy" but there's an oxford comma later on in the docs, so we should probably use one here also.

docs/build/usage.md
```diff
@@ -283,6 +283,75 @@ command:
...
+    - the parent message if this message is a nested definition
+    - Custom options used in the enum, enum values, and the file
+      in which the message is defined
+  - Services
+    - request & response types referenced in methods
+    - custom options for the service, its methods, and the file
+      in which the message is defined
+
+As an example, consider the these sample protos:
+```proto
```
doriable, line 318: Does `proto` work? Looking at the other files in the docs, it seems that `protobuf` is the language setting for the code block.

I like the example though, makes the behaviour very clear :)

robbertvanginkel: Proto did work for me locally, but I've updated to `protobuf` to align with the others.

docs/build/usage.md
```diff
@@ -283,6 +283,75 @@ command:
 $ buf build --path path/to/foo.proto --path path/to/bar.proto
 ```
 
+## Limit to specific types
+
+The output `FileDescriptorSet` or image from a `buf build` will contain all types
+declared in the module. For advanced use cases, users may want an image or
+FileDescriptorSet containing only a subset of types. Introduced in `v1.1.0`, the
```
doriable, line 290: Backticks for `FileDescriptorSet` ^^"

doriable (approved):
Tiny nit, but stamping early, as I agree with the content and don't want to block. @lucperkins should have the final say though :)

lucperkins:
I have some desired tech-writer-ly changes but I can push that through in a follow-up PR. LGTM.

lucperkins (approved):

docs/build/usage.md
robbertvanginkel, line 290:
```suggestion
`FileDescriptorSet` containing only a subset of types. Introduced in `v1.1.0`, the
```

PR #119, opened 2022-03-01
Rework type option explanation

lucperkins:
This PR applies a tech-writer-y veneer to #118.

### DIFF
--- a/docs/build/usage.md
+++ b/docs/build/usage.md
@@ -285,38 +285,37 @@ $ buf build --path path/to/foo.proto --path path/to/bar.proto
 
 ## Limit to specific types
 
-The output `FileDescriptorSet` or image from a `buf build` will contain all types
-declared in the module. For advanced use cases, users may want an image or
-`FileDescriptorSet` containing only a subset of types. Introduced in `v1.1.0`, the
-`--type` flag accepts fully qualified Protobuf names and limits the output image
-to contain only descriptors required to represent those types and their required
-dependencies. The type flag supports messages, enums, and services.
+When you run `buf build` to create a [`FileDescriptorSet`][filedescriptorset] or Buf [image], the
+output contains all of the types declared in the [module]. But for some advanced use cases, you may
+want the image or `FileDescriptorSet` to contain only a subset of the types.
+
+Versions 1.1.0 and later of the `buf` CLI include a `--type` option for the `buf build` command that
+enables you to supply a fully qualified Protobuf name and limit the resulting image or
+`FileDescriptorSet` to only the descriptors required to represent those types and their required
+dependencies. This example usage restricts the types to `pkg.foo.Bar`:
 
 ```terminal
 $ buf build --type pkg.foo.Bar
 ```
 
-A descriptor is said to require another descriptor if the dependent descriptor
-is needed to accurately and completely represent that descriptor. These include:
-  - Messages
-    - messages & enums referenced in fields
-    - proto2 extension declarations for this field
-    - the parent message if this message is a nested definition
-    - custom options for the message, its fields, and the file in which the
-      message is defined
-  - Enums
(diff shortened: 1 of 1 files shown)


docs/build/usage.md
```diff
@@ -285,38 +285,42 @@ $ buf build --path path/to/foo.proto --path path/to/bar.proto
...
+  - Any [proto2] extension declarations for message fields
+  - The parent message if this message is a nested definition
+  - Any custom options for the message, its fields, and the file in which the message is defined
+- [Enums], including:
+  - The enum value descriptors for this enum
+  - The parent message is this enum is a nested definition
+  - Any custom options used in the enum, enum values, and the file in which the enum is defined
+- [Services], including:
+  - Request and response types referenced in service methods
+  - Any custom options for the services, its methods, and the file in which the service is defined
```
robbertvanginkel, line 317: Here the `including` parts don't really relay well the message I initially wanted to get across. The goal is to describe what counts as a "required dependency" in the part above:
>  only those descriptors required to represent those types and their required
dependencies

I think now it looks more like these sub bullets refer to the "including".

lucperkins: @robbertvanginkel I'm still struggling a bit with the wording here, but I made another pass at this.

docs/build/usage.md
```diff
@@ -285,38 +285,55 @@ $ buf build --path path/to/foo.proto --path path/to/bar.proto
...
-  - Services
-    - request & response types referenced in methods
-    - custom options for the service, its methods, and the file
-      in which the message is defined
-
-As an example, consider the these sample protos:
-```protobuf
---- foo.proto ---
+Three Protobuf constructs can act as descriptors: [messages], [enums], and [services]. Each of these
+constructs has a set of dependent descriptors that are included in the build:
```
robbertvanginkel, line 304:
```suggestion
The type flag accepts fully qualified names for [messages], [enums], and [services]. The following qualify as dependent descriptors that are included in the build:
```

robbertvanginkel: There are more than 3 descriptors, they're just not supported here as top level arguments. How about this?

docs/build/usage.md
```diff
@@ -285,38 +285,55 @@ $ buf build --path path/to/foo.proto --path path/to/bar.proto
...
+
+- [Messages]
+  - Messages and enums referenced in message fields
+  - Any [proto2] extension declarations for message fields
+  - The parent message if this message is a nested definition
+  - Any custom options for the message, its fields, and the file in which the message is defined
+- [Enums]
+  - The enum value descriptors for this enum
+  - The parent message is this enum is a nested definition
+  - Any custom options used in the enum, enum values, and the file in which the enum is defined
```
robbertvanginkel, line 314:
```suggestion
  - Any custom options for the enum, enum values, and the file in which the enum is defined
```
consistency with the other 2.

docs/build/usage.md
```diff
@@ -285,38 +285,55 @@ $ buf build --path path/to/foo.proto --path path/to/bar.proto
...
+You can specify multiple types by applying the `--type` option multiple times, as in this example:
+
+```terminal
+$ buf build \
+  --type acme.weather.v1.Units \
+  --type acme.weather.v1.CurrentWeather.Temperature
+```
+
+In this case, all dependent descriptors are built for both `acme.weather.v1.Units` and
+`acme.weather.v1.CurrentWeather.Temperature`.
```
robbertvanginkel, line 329:
```suggestion
In this case, dependent descriptors for both `acme.weather.v1.Units` and
`acme.weather.v1.CurrentWeather.Temperature` are included in the output.
```
Nit: This feature will still build a whole module, it just restricts the output of the build.

lucperkins:
@robbertvanginkel I think this should be good for now, but let's revisit if this proves to be unclear for users.

PR #122, opened 2022-03-03
Remove "very" from the docs

lucperkins:
- Remove usage of very
- Fix misspelling in comment


vale/Buf/very.yml
```diff
@@ -0,0 +1,7 @@
+# Lint rule for keeping us from using "very," the laziest of all adjectives
```
doriable, line 1: 😹

doriable (approved):

vale/Buf/very.yml
lucperkins: Where's the lie?!?!

PR #123, opened 2022-03-04
BSR npm registry docs

lucperkins:
This is still a bit on the rough side and I haven't been able to personally vet some of these things, so please feel free to scrutinize as harshly as you need to.

Preview: https://deploy-preview-123--buf.netlify.app/bsr/remote-generation/npm


docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
+---
+id: npm
+title: The BSR npm registry
+---
+
+> Remote code generation is an **alpha feature**. We started with [Go](../../tour/use-remote-generation.md) and have since added support for generated JavaScript and TypeScript code stubs. We intend to add support for other languages as well. [Let us know](/contact.md) which language we should tackle next.
+
+Now that the BSR supports **remote code generation**, you no longer have to maintain Protobuf files, `protoc`-based plugins or generate code locally. This is especially useful for API clients, who just want a [Go] SDK to start consuming an API immediately.
+
+First offering: Go module proxy
```
johanbrandhorst, line 10: Is this right?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+Why it was tricky: the npm registry needs to rewrite imports from local file paths to absolute paths
+
+You no longer have to maintain Protobuf files, runtime library dependencies, and protoc-based plugins. And most importantly, you no longer need to generate code locally.
+
+Implements the npm's [public registry API][registry]
```
johanbrandhorst, line 15: Never seen it called "the npm" before, but I guess if I read it out loud it makes sense. Should npm be capitalized?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+It generates packages on the fly when you `npm install`.
+
+## Package names
+
+import Syntax from "@site/src/components/Syntax";
+
+<Syntax
+  title="Name syntax for BSR npm registry packages"
+  examples={["@buf/grpc_web_googleapis_googleapis"]}
```
johanbrandhorst, line 24: We may want to use the `protocolbuffers/js` template instead, the grpc-web template doesn't make that much sense for the `googleapis` specifically since you can't expect to talk gRPC-web to GCP.

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+
+This binds the scope `@buf` to the Buf registry and updates your global [`.npmrc`][npmrc]. With this set, you can install `@buf/*` packages in any npm project with a `package.json` configuration.
+
+```terminal
+$ npm install @buf/grpc_web_googleapis_googleapis
+```
+
+### Time to install
+
+You may notice that installing packages from the BSR npm registry using `npm install` typically takes longer than installing from the default public npm registry at https://registry.npmjs.org. That's the case because packages are generated "on the fly," that is, packages aren't pre-built
```
johanbrandhorst, line 57: Full stop?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+
+We provide more details around this [below](#how)
```
johanbrandhorst, line 59: Full stop?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Other package managers
+
+Because the Buf npm registry implements npm's [public registry API][registry], you should be able to use it with package management tools outside of npm, such as [Yarn] and [pnpm], though with [some known limitations](#yarn).
```
johanbrandhorst, line 63: Passive voice?
```suggestion
Because the Buf npm registry implements npm's [public registry API][registry], you can use it with package management tools outside of npm, such as [Yarn] and [pnpm], though with [some known limitations](#yarn).
```

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+
+## Generating private packages {#generate}
+
+To generate npm packages from private Buf [modules], you need to configure npm to send an authentication token with each request to the BSR npm registry. Add a line with this syntax to your [`.npmrc`][npmrc] file:
+
+<Syntax
+  title="npmrc token syntax"
+  examples={["//npm.buf.build/a75c6ca7a..."]}
```
johanbrandhorst, line 71: According to https://docs.npmjs.com/using-private-packages-in-a-ci-cd-workflow, this should probably be
```suggestion
  examples={["//npm.buf.build/:_authToken=a75c6ca7a..."]}
```
. Do we want to reference this page in our docs?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+  segments={[
+    {separator: "//"},
+    {label: "npm.buf.build", kind: "static"},
+    {separator: "/"},
+    {label: "token", kind: "variable"},
```
johanbrandhorst, line 76: Presumably this should be rewritten a bit too.

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+  ]}
+/>
+
+You can use an existing auth token or generate a new one. To create a new one, log into the [BSR], navigate to your [user settings][settings] page, and click **Create Token**.
+
+## Restrictions
+
+### Runtime dependencies
+
+You can use [labels] to declare runtime dependencies for plugins. The BSR npm registry currently supports [semantic versioning][semver] for versions but _not_ semver [ranges] like `>=1.2.7` or `<1.3.0`.
```
johanbrandhorst, line 86: We can include a reference to the "authoring a plugin" section here to have an easy way for users to discover more.

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+
+```javascript
+// generated import
+require("../../google/storage/v1/storage_pb.js");
+
+// replacement
+require("@buf/grpc_web_googleapis_googleapis/google/storage/v1/storage_pb.js");
+```
+
+Why it was tricky: the npm registry needs to rewrite imports from local file paths to absolute paths
```
johanbrandhorst, line 116: Full stop?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+
+### Correctness
+
+All generated code must be syntactically correct.
+
+## How the npm registry works {#how}
+
+When you request a package for a Buf [module][modules] using `npm install` or an analogous method, the BSR npm registry runs the Protobuf [plugins] specified in the module's [template].
+
+> TODO: diagram
```
johanbrandhorst, line 104: Is this ready?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,136 @@
...
+require("../../google/storage/v1/storage_pb.js");
+
+// replacement
+require("@buf/grpc_web_googleapis_googleapis/google/storage/v1/storage_pb.js");
+```
+
+Why it was tricky: the npm registry needs to rewrite imports from local file paths to absolute paths
+
+Implements the npm's [public registry API][registry]
+It generates packages on the fly when you `npm install`.
```
johanbrandhorst, line 119: These read like leftover notes.

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,153 @@
...
+To install npm packages generated from private [Buf modules][modules], you need to configure npm to send an authentication token with each request to the BSR npm registry. Add a line with this syntax to your [`.npmrc`][npmrc] file:
+
+<Syntax
+  title="npmrc token syntax"
+  examples={["//npm.buf.build/84612b6cbe8f41ddb3ac46c691f05a978e20a555ba5d4bc289a0a8289abe75ef"]}
+  segments={[
+    {separator: "//"},
+    {label: "npm.buf.build", kind: "static"},
+    {separator: "/"},
+    {label: "token", kind: "variable"},
```
timostamm, line 75: We're showing the syntax as 

```
//npm.buf.build/{token}
```

But it should be

```
//npm.buf.build/:_authToken={token}
```

```suggestion
  title="npmrc token syntax"
  examples={["//npm.buf.build/:authToken=84612b6cbe8f41ddb3ac46c691f05a978e20a555ba5d4bc289a0a8289abe75ef"]}
  segments={[
    {separator: "//"},
    {label: "npm.buf.build", kind: "static"},
    {separator: "/:authToken="},
    {label: "token", kind: "variable"},
```

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,153 @@
...
+
+With your npm config set, you can install `@buf/*` packages in any standard npm project with a `package.json` configuration. Here's an example installation command:
+
+```terminal
+$ npm install @buf/protocolbuffers_js_acme_paymentapis
+```
+
+:::warning Slow installation?
+You may notice that installing packages from the BSR npm registry using `npm install` can take longer than installing from the standard npm registry. This happens because packages are generated "on the fly"—that is, they're built upon request and then cached. The first `npm install` typically takes a while, but future `npm install`s should be more brisk. More details are [below](#how-it-works).
+:::
```
timostamm, line 30: This renders as a red box, which looks a bit like a warning ahead of a dangerous command, but with recent changes, installation of the (_really_ huge) googleapis/googleapis seems to take just under a minute for the first time: 

```
➜  npmx npm install @buf/protocolbuffers_js_acme_paymentapis

added 3 packages, and audited 4 packages in 59s
```

And about half the second time:

```
➜  npmy npm install @buf/protocolbuffers_js_acme_paymentapis

added 3 packages, and audited 4 packages in 27s
```

That's not great, but maybe we can get away with an orange warning or an alternative that doesn't suggest danger?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,153 @@
...
+
+You can use an existing auth token or generate a new one. To create a new one, log into the [BSR], navigate to your [user settings][settings] page, and click **Create Token**.
+
+## Other package managers
+
+Because the Buf npm registry implements npm's [public registry API][registry], you should be able to use it with package management tools outside of npm, such as [Yarn] and [pnpm], though with [some known limitations](#yarn).
+
+## Restrictions
+
+The BSR npm registry has some known limitations that you should be aware of.
```
timostamm, line 87: "Known limitations" sounds so much better than "Restrictions" :raised_hands:

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,153 @@
+---
+id: npm
+title: BSR npm registry
```
timostamm, line 3: The title for the Go module proxy is "Consume generated Go code". This one seems much more direct and understandable. I wonder if we might want to rename the Go module proxy page?

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,153 @@
...
+
+[Yarn] versions greater than [v1.10.0][yarn_v1] and less than [v2.0.0] are _not_ supported. These versions of Yarn require the `shasum` field in the dist object to be set, but the BSR can't compute a digest without generating the code for all possible versions of the package.
+
+### Import rewrites
+
+The BSR rewrites all `import` statements and `require()` calls in `.js` and `.d.ts` files, matching generated files with their `.proto` sources based on their path. Generated files _must_ use the same path as their `.proto` counterparts, although suffixes like `_pb` and additional file extensions are allowed.
+
+## How it works
+
+When you install a package for a Buf [module][modules] using `npm install`, [`yarn install`](#yarn), or an analogous method, the BSR npm registry runs the Protobuf [plugins] specified in the module's [template]. Once generation is complete, the BSR returns the generated assets to your npm client, which finally writes those assets to your `node_modules`.
```
timostamm, line 103: Do you think people will be interested in the inner workings? I only added "How it works" in the draft as a preface to the limitations, especially for the import rewrite logic, which relies on generated file paths to match source file paths (but ignores extensions). Please drop or merge with "Import rewrites" as you see fit.

docs/bsr/remote-generation/npm.md
lucperkins: Yeah, we probably should. I'll address that in a separate PR.

docs/bsr/remote-generation/npm.md
lucperkins: Well, this currently _is_ orange, but I've changed this to an "info" block, as we're not really "warning" per se but rather setting expectations

docs/bsr/remote-generation/npm.md
lucperkins: Okay, I've merged, but I think I'm maybe still not quite clear on what the implications are for users, so definitely have a look at the merged version

timostamm: Good call on clarity.

_Yarn compatibility_ is the only limitation for users who want to install a generated package. 

_Runtime dependencies_ is relevant for plugin authors. They might expect to be able to give a semver range.

_Import rewriting_ is relevant for plugin authors. Let's say we have a proto file `a.proto` in module `a` that imports `b.proto` from module `b`. Generating the package `@buf/a` will only generate the files of module `a`, and we want the import statement in `@buf/a/a.js` to point to the other module / package `@buf/b/b.js`.
But code generators don't know about modules. The generated `a.js` will try to import `b.js`, which does not exist in `@buf/a`. The BSR will see the import for `b.js`, deduce that this was an import for `b.proto`, and rewrite the import to `@buf/b/b.js`. For deducing which import belongs to which module, the BSR matches import paths with proto paths, for example `foo/bar.js`, `foo/bar_pb.js`, `foo/bar_grpc.d.ts` are all recognized as derivates of `foo/bar.proto`.
Plugin authors should be aware that if they generate for example `foo-bar.js` for `foo/bar.proto`, the BSR will be unable to rewrite imports, and modules with dependencies won't work.

lucperkins: Okay, I've added a table that may help to present this. I also intend to add an example project in `buf-examples` later on. Let me know if the table is correct and/or helpful.

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,143 @@
...
+
+Because the Buf npm registry implements npm's [public registry API][registry], you should be able to use it with package management tools outside of npm, such as [Yarn] and [pnpm], though with [some known limitations](#yarn).
+
+## Known limitations
+
+The BSR npm registry has a few limitations that you should be aware of.
+
+### Yarn compatibility {#yarn}
+
+[Yarn] versions greater than [v1.10.0][yarn_v1] and less than [v2.0.0] are _not_ supported. These versions of Yarn require the `shasum` field in the dist object to be set, but the BSR can't compute a digest without generating the code for all possible versions of the package.
```
timostamm, line 93: `[v2.0.0]` isn't a link, turns out v2.0.0 was never released. https://github.com/yarnpkg/berry hosts the post-v1 Yarn, but they skipped v2 despite doing some pre-releases and went with v3. What a mess.

I suggest we just use `[v2](https://github.com/yarnpkg/berry)` here, so it doesn't look like there was a huge gap we don't support. I just tested an install with yarn v3.2.0 and it went through.

docs/bsr/remote-generation/npm.md
```diff
@@ -0,0 +1,143 @@
...
+// replacement
+require("@buf/grpc_web_googleapis_googleapis/google/storage/v1/storage_pb.js");
+```
+
+What this means in practice is that files generated by Protobuf plugins _must_ use the same path as their `.proto` counterparts, although suffixes like `_pb` and additional file extensions are allowed. The table below shows an original Protobuf filepath (`foo/bar.proto`) and which generated filepaths would be acceptable (or not).
+
+Proto filepath | Path of generated file | Acceptable?
+:--------------|:-----------------------|:-----------
+`foo/bar.proto` | `foo/bar.js` | ✅
+`foo/bar.proto` | `foo/bar_pb.js` | ✅
```
timostamm, line 116: Much better than any wall of text I'd have come up with :raised_hands:

timostamm (approved):
IMO this gives a solid overview and reads well, and just needs a fix for the yarn links.

docs/bsr/remote-generation/npm.md
lucperkins: What a mess indeed!

docs/bsr/remote-generation/npm.md
lucperkins: Good! Glad this clears it up a bit

lucperkins:
@timostamm Cool cool! I'm gonna wait to merge this until we're ready to launch.

PR #134, opened 2022-03-15
Modify remote url in "list remote" doc

rubensf:
Github does not accept git:// anymore - the command as is returns:

Failure: could not clone git://github.com/bufbuild/buf-tour.git: exit status 128
fatal: remote error:
  The unauthenticated git protocol on port 9418 is no longer supported.
Please see https://github.blog/2021-09-01-improving-git-protocol-security-github/ for more information.


lucperkins:
@rubensf Good catch

PR #142, opened 2022-03-23
Fix typo in buf.lock

craigpastro:
Just a small fix to the introduction to buf.lock which I noticed this morning.


lucperkins:
Good catch. Thanks!

PR #145, opened 2022-03-24
Clarify JS + TS support

lucperkins:
This PR updates our remote code gen documentation for npm-based workflows. It adds a section on package versions and provides some other clarifications not present in the current doc.

Preview:

https://deploy-preview-145--buf.netlify.app/bsr/remote-generation/js

### DIFF
--- a/docs/bsr/remote-generation/js.md
+++ b/docs/bsr/remote-generation/js.md
@@ -39,7 +39,7 @@ Template | What it generates
 [`protocolbuffers/js`][pb-js] | JavaScript code stubs (`.js`)
 [`grpc/web`][grpc-web] | JavaScript code stubs (`.js`), TypeScript type definitions (`.d.ts`)
 
-In addition to these official templates, you can use any templates you like with npm&mdash;including templates that [you upload yourself](../remote-generation/template-example.md)&mdash;provided that those templates generate valid JavaScript and/or TypeScript code stubs. Templates that generate invalid JavaScript/TypeScript or other languages aren't supported. This operation, for example, would fail because the template, [`protocolbuffers/go`][pb-go], isn't supported:
+In addition to these official templates, you can use any templates you like with npm&mdash;including templates that [you upload yourself](../remote-generation/template-example.md)&mdash;provided that those templates generate valid JavaScript and/or TypeScript declaration files. Templates that generate invalid code aren't supported. This operation, for example, would fail because the template, [`protocolbuffers/go`][pb-go], isn't supported:
 
 ```terminal
 $ npm install @buf/protocol_buffers_go_acme_paymentapis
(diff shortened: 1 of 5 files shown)


docs/bsr/remote-generation/js.md
```diff
@@ -24,28 +24,55 @@ $ npm config set @buf:registry https://npm.buf.build
...
 ## Installing packages {#install}
 
-With your npm config set, you can install `@buf/*` packages in any standard npm project. Here's an example installation command:
+With your npm config [set](#setup), you can install `@buf/*` [packages](#package-names) in any standard npm project. Here's an example installation command:
 
 ```terminal
 $ npm install @buf/protocolbuffers_js_acme_paymentapis
 ```
 
+You can use only the [supported templates](#templates) with npm. This operation, for example, would fail because the template, [`protocolbuffers/go`][pb-go], isn't supported:
```
timostamm, line 50: This might be misunderstood. It's perfectly fine for users to create their own templates, or even their own plugins, and run it through our npm registry. 

But the npm registry expects valid JavaScript and/or TypeScript declaration files. If you try to npm install a golang template, our registry will try to parse Go files as JavaScript, and fail with a syntax error.

lucperkins: Got it. I'll update this.

docs/bsr/remote-generation/js.md
```diff
@@ -24,9 +24,33 @@ $ npm config set @buf:registry https://npm.buf.build
...
+## Available templates {#templates}
+
+The table below lists the generation [templates](overview.md#templates) that are officially supported by the BSR:
+
+Template | What it generates
+:--------|:-----------------
+[`protocolbuffers/js`][pb-js] | JavaScript code stubs (`.js`)
+[`grpc/web`][grpc-web] | JavaScript code stubs (`.js`), TypeScript type definitions (`.d.ts`)
+
+In addition to these official templates, you can use any templates you like with npm&mdash;including templates that [you upload yourself](../remote-generation/template-example.md)&mdash;provided that those templates generate valid JavaScript and/or TypeScript code stubs. Templates that generate invalid JavaScript/TypeScript or other languages aren't supported. This operation, for example, would fail because the template, [`protocolbuffers/go`][pb-go], isn't supported:
```
timostamm, line 42: Isn't _TypeScript code stubs_ a bit too ambiguous?

We really only support _TypeScript declaration files_. Their purpose is specifically to provide strong typing for `.js` files. The `.js` files can be consumed from a package by node.js and all bundlers, while the TypeScript compiler uses the `.d.ts` files to check types (a bit similar to C headers). 


```suggestion
In addition to these official templates, you can use any templates you like with npm&mdash;including templates that [you upload yourself](../remote-generation/template-example.md)&mdash;provided that those templates generate valid JavaScript and/or TypeScript declaration files. Templates that generate invalid code aren't supported. This operation, for example, would fail because the template, [`protocolbuffers/go`][pb-go], isn't supported:
```

timostamm (approved):

docs/bsr/remote-generation/js.md
lucperkins: Yeah, that's a fair distinction. Force of habit 😄

PR #162, opened 2022-06-13
minor updates to buf tour

pkwarren:
Fix a few typos and update the service name to match the expected state
at the point in the tour when we add the PurchasePet method.

### DIFF
--- a/docs/tour/use-a-workspace.md
+++ b/docs/tour/use-a-workspace.md
@@ -9,7 +9,7 @@ That's a great start, but product requirements always evolve and new features ne
 time.
 
 In this section, you'll incorporate another dependency into your `PetStoreService` API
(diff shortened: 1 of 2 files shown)


docs/tour/use-a-workspace.md
```diff
@@ -216,7 +216,7 @@ Adapt the `PetStoreService` with the `PurhcasePet` endpoint like this:
 +
 +message PurchasePetResponse {}
 +
- service PetStore {
+ service PetStoreService {
```
pkwarren, line 219: At this point in the tour, you've already renamed to fix the lint error.

mfridman: Good catch

mfridman (approved):

PR #164, opened 2022-06-25
Update highlighting to match diff

SyrieBianco:
Fixes TCN-163

The current version of this diff just updates the manual highlighting to be accurate, though we could replace the manual highlighting with the `diff` language rather than the `yaml` key. There is a slight formatting change. 

For now, I'm just going to update it to be consistent, but we should consider if we can deal with the `diff` formatting so as to not manually maintain line highlighting

<details>
<summary>Expand to see formatting differences</summary>
1. yaml formatting with manual Highlighting
`yaml title="buf.yaml" {6-7}`
Screen Shot 2022-06-24 at 8 23 13 PM

2. diff formatting (without yaml syntax highlighting)
`diff title="buf.yaml" ` 
Screen Shot 2022-06-24 at 8 13 00 PM
</details>

### DIFF
--- a/docs/tour/add-a-dependency.md
+++ b/docs/tour/add-a-dependency.md
@@ -29,7 +29,7 @@ $ rm -rf google
 
 Now remove the `google/type/datetime.proto` reference from your [`buf.yaml`](../configuration/v1/buf-yaml.md):
 
-```yaml title="buf.yaml" {5-6}
+```yaml title="buf.yaml" {6-7}
  version: v1
  name: buf.build/$BUF_USER/petapis
  lint:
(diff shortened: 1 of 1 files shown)


docs/tour/add-a-dependency.md
```diff
@@ -29,7 +29,7 @@ $ rm -rf google
 
 Now remove the `google/type/datetime.proto` reference from your [`buf.yaml`](../configuration/v1/buf-yaml.md):
 
-```yaml title="buf.yaml" {5-6}
+```yaml title="buf.yaml" {6-7}
```
SyrieBianco, line 32: Checking to see if switching to `diff` takes care of the highlighting for us in the netlify deploy (though there may be a reason we're not using that here)

```suggestion
```diff title="buf.yaml" 
```

docs/tour/add-a-dependency.md
```diff
@@ -29,7 +29,7 @@ $ rm -rf google
 
 Now remove the `google/type/datetime.proto` reference from your [`buf.yaml`](../configuration/v1/buf-yaml.md):
 
-```yaml title="buf.yaml" {5-6}
+```diff title="buf.yaml"
```
SyrieBianco, line 32:
```suggestion
```yaml title="buf.yaml" {6-7}
```

timostamm (approved):

PR #166, opened 2022-07-01
Vercel migration: 20mins

paul-sachs:
Fixes OPS-94


timostamm:
Are we switching to ease some pain points with Netlify? Would be nice to have some info on this PR for posterity :)

paul-sachs:
@timostamm will have more details when I merge this but the short version is that Netlify increased their prices 10x. This is me messing with switching to vercel instead.

timostamm (approved):
> Netlify increased their prices 10x

:panic_stamp:  :grinning:

paul-sachs:
Fixes OPS-94

PR #168, opened 2022-07-15
Missing directory according to the example

unmultimedio:
If I'm understanding correctly, in the example, this should be generated within `gen/proto/go`, right? https://docs.buf.build/configuration/v1/buf-gen-yaml#out


mfridman (approved):
might be worth testing. But the sentence above does say `gen/proto/go`

and also reading the `out:` from the gen file checks out. Great spot 🎉

PR #178, opened 2022-08-06
remote generation: fix invalid template example

twilly:
In the remote generation docs there's an example of an invalid npm fetch
of a Go module. This makes sense; Go isn't JavaScript.

However, the example shows an different error than described. Using four
underscores is illegal and causes an error in
`bufnpmdpackage/newPackageNameFromString` rather than a "this isn't a
JavaScript template/package" error.

I've corrected the npm package name for the correct "protocolbuffers"
template owner and updated the npm response. Interestingly the new error
is from npm (EINVALIDPACKAGENAME) and not that the generated code isn't
for JavaScript.


timostamm (approved):
Thanks for catching that, Tristan!

> Interestingly the new error is from npm (EINVALIDPACKAGENAME) and not that the generated code isn't for JavaScript.

We can't really tell whether the generated output is suitable for an npm package. If I'm not mistaken, we will soon have a feature that lets plugins declare which registries they are suitable for. This will allow us to have a meaningful error message here.

PR #180, opened 2022-08-15
Replace Buf's TM with R in the title

seankimdev:
Closes FE-347

Before:
image

After:
image

PR #181, opened 2022-08-15
Change TM in titles to (R)

stefanvanburen:
Noticed this looking through the docs today - similar to #179. ™️ to ®. (There are two emojis - ® or ®️ - using the former, see below for a comparison).

For example, currently we have this in Safari:

image

With the former:

image

With the latter:

image


paul-sachs (approved):

seankimdev:
Wow we updated this at the exact same time 😅 I'll close my PR https://github.com/bufbuild/docs.buf.build/pull/180.
Thanks for updating it!

stefanvanburen:
@seankimdev 😅 sorry about that, should have checked Linear beforehand to see if you all had it tracked

PR #185, opened 2022-08-22
Add Machine role docs

stefanvanburen:
This probably shouldn't go out until we have the frontend piece deployed; just getting this queued up.

Definitely appreciate any wordsmithing and suggestions on placement within the page - I think it looks a little out of place currently, but I think we _do_ want to give readers a strong explanation for what the role is used for.


docs/bsr/user-management.md
```diff
@@ -21,7 +21,13 @@ Every user that is part of an organization has an explicit role. Note that users
...
 
 - Can view the organization and its members.
-- Have the [Base resource role](#base-resource-roles) over the organizations resources, which defaults to [Write](#write).
+- Have the [Base resource role](#base-resource-roles) over the organizations resources, which defaults to [Write](#resource-write).
+
+### Machine {#org-machine}
+
+- Can write to organization resources, regardless of the organization's [base resource roles](#base-resource-roles)
+
+This role is useful in CI pipelines - you can set the organization base roles to **Read** and configure a Machine user to write to a BSR repository on merge, for example.
```
stefanvanburen, line 30: I initially had something like this as a bullet under "Resource roles" (alongside the bullets starting "Outside collaborators" and "Elevated permissions"), but I'm not sure it belongs there?

I think it doesn't feel quite right in those bullets because the preceding paragraph says:

> In some situations, however, you'll need to give additional permissions to individual users over a user- or organization-owned resource.

I feel like "you'll need to give additional permissions to individual users" doesn't  seem quite right in the context of the machine role. It's more like "modifying permissions of your regular users and giving permissions specifically to particular users to perform actions on your organization's behalf", or something.

cyinma, line 30: can we link the `**Read**` to `bsr/user-management#resource-read`?

stefanvanburen: updated! PTAL

cyinma (approved):

PR #191, opened 2022-09-03
Fix some minor formatting issues

rebello95:
Fixes a couple of minor formatting issues in the documentation.


docs/tour/use-a-workspace.md
```diff
@@ -216,6 +216,9 @@ Adapt the `PetStoreService` with the `PurchasePet` endpoint like this:
 +
 +message PurchasePetResponse {}
 +
+
+ ...
+
```
stefanvanburen, line 221: with this added, I think you'll need to change the line highlight for the codeblock on line 200; it looks like it should be `26` instead of `23` now? (alternatively, could just nix this change and assume that the code was meant to be added right above the `service` declaration 😄)

stefanvanburen (commented):
hi @rebello95, thanks for the contribution! one minor nit but generally looks good 😃

stefanvanburen (approved):
lgtm, thanks @rebello95!

PR #192, opened 2022-09-06
Add note about updating dependencies first

stefanvanburen:
If someone depends on a module with a dependency on googleapis, they'll need to update the dependency first. "Upstream" is the language we use over in the tour.


rubensf:
Approved

PR #194, opened 2022-09-13
Remote generation changes feature branch

srikrsna-buf:
- #195
- #197
- #196
- #198

Also:

- ~update the `bsr/remote-generation/overview` section~ 
  - ~going to remove this entirely and revamp the bsr/{introd, overview,usage} sections~
- ~migration guides?~ added in this PR
- ~Remove or deprecate the authoring template|plugin sections~
- ~link old go/npm sections to new go/npm sections~
- ~add blurb on old go/npm sections. Deprecated, sunset, legacy, etc.~

- [ ] Update `tour/use-remote-generation` with new Go module proxy and remove sections that don't make sense, like synthetic versions


srikrsna-buf:
As a consumer, I think the proto plugin section, particularly the import rewrite section, is a bit confusing and out of context. While accurate, I think this is suited for plugin authors more than consumers. 

More than promoting `connect-web` and `protobuf-es`, this is one of the reasons I don't want to merge https://github.com/bufbuild/plugins/pull/81. The need to support two different forms of generation and explaining to consumers the difference and also highlighting which plugin outputs the code in which form is very confusing to a consumer. Moreover the project is on [minimal support](https://github.com/protocolbuffers/protobuf-javascript#project-status-as-of-september-2022). In this state I would prefer generating code plugin wise and avoid this section altogether. We can either do this by not supporting generating assets for the grpc-web, js, and grpc-node (my preference as they can always migrate to `buf generate module`) or updating the plugins to output to separate packages. This will be inline with how the `goproxy` works. @timostamm would like to hear your thoughts on this.

With that being said we still have to explain the generated code for each plugin, I prefer doing this at a plugin level. The same holds true for `go` plugins, it is best if we explain how various plugins output code and it should be done at the plugin level and should ideally remain outside our docs.

timostamm:
> @timostamm would like to hear your thoughts on this.

I'm fairly sure that the NPM registry docs are overwhelming for the majority of users.  :sweat:

In general, I think we're not really highlighting the benefits enough yet. The registry makes generated code available like any other dependency: Zero tooling to install; no need to keep generator and supporting library versions in sync; dependabot and others will pick it up; API producers can provide two lines of installation instructions instead of maintaining an SDK. These are genuinely useful features!

But to the crux: 

If we want to tell a compelling and cohesive story, curating only the plugins protoc-gen-es, protoc-gen-connect-web and protoc-gen-connect-node is our best option IMO. The package splitting could be explained in a "How it works" section at the bottom of the docs, our plugin readme's could give concrete examples, and in-depth explanation could live in the [@bufbuild/protoplugin readme](https://github.com/bufbuild/protobuf-es/tree/main/packages/protoplugin). This won't give power users all the knobs and features, but it will be a complete story without rough edges and head-scratching.

If we curate protoc-gen-js, protoc-gen-grpc-web etc., I don't see a clear path that avoids leaking the complexity to users. Some plugins will only generate CommonJS, some only ECMAScript modules, some only generate JS, some generate TS or type declarations. The distinction between protoc-gen-js and protoc-gen-es will be mysterious, and so on. I don't think it's viable to fork all generators to support import rewriting - it only solves part of the problem, but at a high cost of maintenance and adding even more plugins to the picture.  
In this situation, documenting at the plugin level is compelling, but where would that documentation actually live? I can imagine we add docs to the protoc-gen-connect-web readme, but we cannot reasonably ask the grpc-web folks to update [their readme](https://github.com/grpc/grpc-web#code-generator-plugin).

srikrsna:
> If we curate protoc-gen-js, protoc-gen-grpc-web etc., I don't see a clear path that avoids leaking the complexity to users.

I am in favor of not making them a part of `npm` registry. But they will still be hosted. Users can use `buf generate` to invoke them.

> In this situation, documenting at the plugin level is compelling, but where would that documentation actually live?

IMO it should live on the plugin's page on the BSR. We can highlight differences and any patches (`grpc-go` and `connect-go`)

> In general, I think we're not really highlighting the benefits enough yet.

Now that you list them yes I see that we can do better. Having not run into some of them you listed I never thought of them as potential problems this solves, but I can totally see that happening. We can maybe do this as part of an overview @mfridman wdyt?

mfridman:
Great discussion @srikrsna @timostamm 

Agreed, the NPM registry docs for import rewrites at the plugin _and_ server level is confusing. I think we can entirely remove this, or create an "Advanced" or "How it works" section to avoid overwhelming users. I'm leaning towards removing this section entirely and adding it back only if enough folks ask about it?

> In general, I think we're not really highlighting the benefits enough yet. The registry makes generated code available like any other dependency: Zero tooling to install; no need to keep generator and supporting library versions in sync; dependabot and others will pick it up; API producers can provide two lines of installation instructions instead of maintaining an SDK. These are genuinely useful features!

Good feedback and suggestions. I'll try to distill this into a paragraph and we can iterate on this.. of all sections it's quite important we get this right.

Re. _which_ plugins work with the BSR npm registry, I think we're in a bit of a disagreement. I'd like to see support for as many plugins as possible, even if they operate slightly differently. Maybe there is an argument for only supporting protoc-gen-{es, connect-web, connect-node} and making _that_ the curated NPM registry experience, which gives us more control. But if we go down this route, we're making this feature _less_ useful for the folks that don't buy into the connect ecosystem.

srikrsna-buf:
> But if we go down this route, we're making this feature _less_ useful for the folks that don't buy into the connect ecosystem.

I don't think this is the connect ecosystem. Having a separate protobuf runtime implementation is well beyond connect. It is the other way around, connect-web and eventually connect-node depend on this new runtime. My reason is not to push connect, but to have a singular clear way for the `npm` registry and protobuf plugins and the runtime in javascript.

JavaScript has long been the place where protobuf was not adopted very well. This is due to the complexity of protoc and also the complexity and minimal support of the runtime. With `protobuf-es` we can really change that. But along with the runtime, other aspects also have to change. One of them being plugin authoring and the idea of plugin dependencies. Having two different runtimes that we support will create a lot of confusion especially when they are fundamentally different. Only supporting `protobuf-es` based plugins will make sure `npm` registry can evolve as needed without the need to support the old runtime and plugins. This will also significantly reduce the overhead of users having to figure out import paths based on plugin.

In addition to that, generating code to the same package is not a complete solution as it doesn't allow a user to choose the version of dependencies which they could in the template based approach. This still remains a problem and something we will need to explain.

mfridman:
Okay, I see your point wrt the two implementations.

> Having two different runtimes that we support will create a lot of confusion especially when they are fundamentally different.

This functionality enables us to support plugins not based on protobuf-es. And thus fallback to the alternate implementation of generating multiple plugins output to the same package.

I don't think it's reasonable for us to maintain / upgrade existing plugins to fit our model.

If I understand correctly, the chain of events will be:

- Buf rolled out a new version of the npm registry
- The old npm registry will be shutdown eventually
- As a user, I was using the grpc/web template
- The new npm registry doesn't support grpc/web, what do I do...?

Given that we _can_ support this user, at the expense of having a different implementation, why wouldn't we? This is the question I keep coming back to, what are we going to tell this user..

Maybe this user was satisfied with same package output, and afaics we don't have a good answer for them.

srikrsna-buf:
> * The new npm registry doesn't support grpc/web, what do I do...?

At this point the user has two options. 
### `connect-web`
`connect-web` does have a `grpc-web` transport to communicate with `grpc-web` backends. The user can switch to `connect-web`, yes it will be a breaking change but definitely for the better. They can plan their switch in such a way that it happens before we sunset the template based registry. I suspect this is what most users will choose if they haven't already.

### `buf generate`
This is an easier switch and the same level of breaking change if we supported the `grpc-web` plugin in the new one. They can use the hosted plugins of `grpc-web` and `js` and point `buf generate` to a commit in the module registry: `buf generate buf.build/acme/petapis:commit`. After that they just have to update their import prefix. This IMO is better than template based npm registry because of not having to deal with semantic versions and understanding what options to configure. Because with these plugins there is no one option fits all approach. As @timostamm pointed out each plugin outputs code in different ways supporting different set of options.


> Given that we _can_ support this user, at the expense of having a different implementation, why wouldn't we?

Because in trying to support this we are increasing the complexity. We have to explain why certain plugins work in one way and others work differently. The whole point of generated assets is that they don't have to worry about such things. Also the question to consumers on which one to pick.

Let's take a similar scenario and ask this question a little differently. I have a working website. Now by adding a different implementation I can support IE. But Microsoft has stopped supporting IE. Now why would I want to support IE? Because we _can_ support it? If so, whenever we try to add a new feature set we have to make sure it is supported in IE or at least bypass IE. This hinders evolution of the website. Our scenario is pretty similar. The project is on minimal support. It has not evolved in a long time.

In any case, at this stage I think we can release without them and if the users really don't like either of the two options (if so we have to explore why) we can enable them before we sunset templates.

mfridman:
> In any case, at this stage I think we can release without them and if the users really don't like either of the two options (if so we have to explore why) we can enable them before we sunset templates.

Okay, there's fairly good points here. I'm okay with closing https://github.com/bufbuild/plugins/pull/81 for now and seeing what the community feedback will be.

Will also update the NPM docs based on the above suggestions.

srikrsna-buf (commented):
There are couple of narrative improvements I think we can make here:

### UX focused 

The docs are very accurate and aptly describe the feature set that we have, but in some places they are more focused on how a problem is being solved and only later do they delve into the user experience. One place where this is seen is the [remote-plugin-execution](https://github.com/bufbuild/docs.buf.build/pull/194/files#diff-be97a81243cab3555804d24a2ebbf33443d95b711e4dc6c94c32f6edd8894c12) docs. 
The introduction reads:
https://github.com/bufbuild/docs.buf.build/blob/7be66a616b34bc0a3c188d9103c4f1dc7f4d24b6/docs/bsr/remote-generation/remote-plugin-execution.md?plain=1#L14

This is accurate and all valid points but the focus is on (important) secondary benefits, the first and most obvious benefit to me is the user experience of it. All they have to do is reference the hosted plugin instead of the local one. It is much later in the docs after introducing all the plugins do we show them an example of the `buf.gen.yaml` using a hosted plugin.

### Simplified names

The names _remote-generation_, _remote-plugin-execution_ are informative and long. No matter how they are actually implemented, from a user's standpoint **Hosted Plugins** just clicks without any explanation. I think we can should change the name for _remote-plugin-execution_ to **Hosted Plugins**. This also ties in to the first point of being UX focused.

**Generated SDKs** definitely is better than remote generation. I am OK with it. The other alternative is **BSR Packages** (similar to GitHub Packages). I would really like to eliminate the word generation from this. More on this in the next section.

### Separating Features

While I know that the package proxies only work because of the Hosted Plugins, we should separate the two features into their own sections under the BSR. We can introduce **Generated SDKs** in the **Hosted Plugins** doc as a way to eliminate the `buf generate` process altogether. 

I say this because the **Generated SDKs** is a feature I want people with no protobuf expertise to use. One I want to market to users as a way to distribute their API both public and internal so that the users even without any protobuf knowledge can just do a `go get` and depend on a module. The API owner can just point to the BSR as the source of documentation and the BSR highlights a way to use the API in the language of their choice.

I don't want to advertise 3 different ways of doing the same thing (`buf generate` with local, `buf generate` with hosted plugins, and `go get`) rather subtly explain how each is different and how one can be better over the other depending on the scenario with distinct target scenarios for each.

ElliotMJackson:
closing... replaced by #220

PR #196, opened 2022-09-13
update docs to use new buf.gen.yaml plugin field

pkwarren:
Update the docs to use the new 'plugin' field which can specify either a local or remote plugin. Update the 'buf.gen.yaml' documentation to include information on the new field. Update the hosted plugins page to document the new plugins.


docs/bsr/remote-generation/hosted-plugins.md
```diff
@@ -23,44 +23,56 @@ With this setup a single developer or thousands of developers at a large organiz
...
-- https://buf.build/grpc/plugins/objc
-- https://buf.build/grpc/plugins/php
-- https://buf.build/grpc/plugins/python
-- https://buf.build/grpc/plugins/ruby
-- https://buf.build/grpc/plugins/web
+The Buf team maintains an open source repository of official plugins at https://github.com/bufbuild/plugins.
+
+### protobuf language plugins
+
+The Buf team has developed tooling to publish all of the plugins built-in to `protoc`. Additionally, other third-party plugins provide language support for languages unsupported by `protoc` (TypeScript, Dart, Swift). Here is a list of supported language plugins:
```
pkwarren, line 34: Updated the documentation here to better reflect that these are not all coming from `protoc` (Dart, Go, etc. are maintained separately).

docs/bsr/remote-generation/hosted-plugins.md
```diff
@@ -23,44 +23,56 @@ With this setup a single developer or thousands of developers at a large organiz
...
+
+- C++: https://buf.build/library/cpp
+- C#: https://buf.build/library/csharp
+- Dart: https://buf.build/library/dart
+- Go: https://buf.build/library/go
+- Java: https://buf.build/library/java
+- JavaScript/TypeScript: https://buf.build/library/protobuf-es
```
pkwarren, line 41: Added protobuf-es.

docs/bsr/remote-generation/hosted-plugins.md
```diff
@@ -23,44 +23,56 @@ With this setup a single developer or thousands of developers at a large organiz
...
+- JavaScript (deprecated): https://buf.build/library/js
+- Kotlin: https://buf.build/library/kotlin
+- Objective C: https://buf.build/library/objc
+- PHP: https://buf.build/library/php
+- Python: https://buf.build/library/python
+- Ruby: https://buf.build/library/ruby
+- Swift: https://buf.build/library/swift
```
pkwarren, line 48: We didn't have Swift listed here before.

docs/bsr/remote-generation/hosted-plugins.md
```diff
@@ -23,44 +23,56 @@ With this setup a single developer or thousands of developers at a large organiz
...
+
+This is powerful because you no longer need to have `protoc` installed, or understand how to invoke it (a daunting task in and of itself). Furthermore you don't need to install additional plugins not already built-in to the `protoc` compiler, such as [protoc-gen-go](https://pkg.go.dev/google.golang.org/protobuf/cmd/protoc-gen-go).
+
+### RPC plugins
```
pkwarren, line 52: Broke this out into separate sections for different types of RPC systems (including Connect and Twirp).

docs/bsr/remote-generation/hosted-plugins.md
```diff
@@ -94,10 +106,10 @@ managed:
   go_package_prefix:
     default: github.com/organization/repository/gen/go
 plugins:
-  - remote: buf.build/protocolbuffers/plugins/go:v1.27.1-1
```
pkwarren, line 97: Not sure if we want to pre-emptively backfill old versions or wait to see which are needed.

docs/bsr/usage.md
```diff
@@ -131,9 +131,9 @@ Add a `buf.gen.yaml` file and list all the plugins as well as their options. Her
 ```yaml title=buf.gen.yaml
 version: v1
 plugins:
-  - name: cpp
+  - plugin: cpp
     out: gen/proto/cpp
-  - name: java
+  - plugin: java
```
pkwarren, line 136: Updated this to be consistent anywhere, however it will be good to note that this only will work on the latest release of the Buf CLI.

docs/bsr/remote-generation/hosted-plugins.md
mfridman: I suggest we use whatever is the latest as of today, and only backfill existing versions if there is a specific ask.

docs/bsr/remote-generation/hosted-plugins.md
```diff
@@ -23,44 +23,56 @@ With this setup a single developer or thousands of developers at a large organiz
...
-- https://buf.build/grpc/plugins/python
-- https://buf.build/grpc/plugins/ruby
-- https://buf.build/grpc/plugins/web
+The Buf team maintains an open source repository of official plugins at https://github.com/bufbuild/plugins.
+
+### protobuf language plugins
+
+The Buf team has developed tooling to publish all of the plugins built-in to `protoc`. Additionally, other third-party plugins provide language support for languages unsupported by `protoc` (TypeScript, Dart, Swift). Here is a list of supported language plugins:
+
+- C++: https://buf.build/library/cpp
```
mfridman, line 36: Do you think we'll want to keep this list around as filler for this page, or will the new `/plugins` page within the BSR be the source of truth. 

In other words, would we just have a blurb along the lines of "To discover hosted plugins please navigate to the Public BSR..."

docs/bsr/usage.md
mfridman: In retrospect, I guess we could have kept `name` / `remote` keys and auto-detected based on the `{ownerName}/plugins/{pluginName}` which endpoint to send the request. 

This would have probably been less disruptive. Nothing to do, just a comment in passing.

docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -101,13 +103,18 @@ on your `PATH`, but is found at `bin/proto/protoc-gen-foo`, you can refer to it
...
     out: gen/foo
     path: bin/proto/protoc-gen-foo
 ```
 
 This field is **exclusive** with `remote` and only works with `name` for local plugins.
 
+#### `revision`
+
+The `revision` is **optional** and may be used along with the `plugin` field to pin an exact version of a remote plugin.
+In most cases, it is recommended to omit the `revision`, in which case the latest revision of the plugin will be used (automatically pulling in the latest bug fixes).
```
mfridman, line 116: Should we add an example here? E.g.,

```
plugins:
  - plugin: buf.build/library/go:v1.28.1
    revision: 1
```

It's nice to have a copy/pastable example right from the docs. (I might have missed it, but didn't see a copyable example containing a  revision)

mfridman (commented):
Great job, this is looking nice 🎉

docs/bsr/remote-generation/hosted-plugins.md
srikrsna-buf: I would point to that or some form of embedding if possible.

srikrsna-buf (commented):
Looks great! Any thoughts on renaming remote plugin execution to Hosted Plugins and moving it out of remote generation?

docs/bsr/usage.md
pkwarren: Discussed this further offline. Even if we had used the existing fields we'd still need to update the Buf CLI to support the new plugins, so perhaps this will work out ok in the end.

docs/bsr/remote-generation/hosted-plugins.md
pkwarren: I'll add that in the overview but I still think it is useful to reference the plugins here for people used to the previous documentation (and also as a reference for what to update their buf.gen.yaml to use).

docs/configuration/v1/buf-gen-yaml.md
pkwarren: Added an example.

docs/bsr/remote-generation/hosted-plugins.md
pkwarren: Added some content describing the main plugins entry page.

docs/bsr/remote-generation/hosted-plugins.md
```diff
@@ -23,52 +23,67 @@ With this setup a single developer or thousands of developers at a large organiz
...
 > To learn more about Buf Plugins check out the [Key concepts documentation](overview.md#concepts).
 
 ## Official plugins
 
-### `protoc`-based plugins
+To discover plugins maintained by the Buf team, go to https://buf.build/plugins.
+This page provides information on all available plugins, including the language type(s) and instructions for use in `buf.gen.yaml`.
 
-The Buf team has developed tooling to automatically sync and publish all of the plugins built-in to `protoc`, which are located under the `protocolbuffers` organization. Here is a list of supported `protoc`-based plugins:
+The source code for official plugins is available at https://github.com/bufbuild/plugins.
```
mfridman, line 33: Suggestion:

"The packaging and distribution source code for Buf-managed plugins is available at..."

Not sure if this reads better though. This sentence makes it seem the plugin source code is at that repository.

docs/bsr/remote-generation/hosted-plugins.md
```diff
@@ -159,20 +174,20 @@ version: v1
...
 ```
 
   </TabItem>
 </Tabs>
 
-Note, we're using the `remote` key instead of `name` to reference a remote plugin, instead of a local one. More information can be [found in the buf.gen.yaml docs](https://docs.buf.build/configuration/v1/buf-gen-yaml#name-or-remote).
+Note, we're using the `plugin` key to reference a hosted plugin. More information can be [found in the buf.gen.yaml docs](https://docs.buf.build/configuration/v1/buf-gen-yaml#plugin-name-or-remote).
 
-> As a best practice, when referencing remote plugins we recommend including the version of the plugin to ensure reproducible code generation.
+> As a best practice, when referencing hsoted plugins we recommend including the version of the plugin to ensure reproducible code generation.
```
mfridman, line 188: Typo `hsoted`

mfridman (approved):
Nits, in a future PR we should remove the demolab example in favor of acme/petapis or bufbuild/eliza. The latter is probably more interesting because there is a real demo API.

docs/bsr/remote-generation/hosted-plugins.md
pkwarren: Done.

PR #197, opened 2022-09-14
Add new npm registry docs

mfridman:
Adding a new section for npm registry. Remaining sections

- [x] add a note on plugin import rewrite feature, such as connect-web / protobuf-es. Not to be confused by server import rewrites by plugins that do not support this. This feature allows outputting plugin dependencies into separate packages, especially useful for common base types.
- [x] figure out how to present / explain plugin versioning schema

Adding this image for discussion.

image


docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,224 @@
...
+} />
+
+In the first example, `@buf/acme_petapis.library_connect-web`, the BSR generates code for the [`acme_petapis`](https://buf.build/acme/petapis) module against the [`library_connect-web`](https://buf.build/library/connect-web) plugin.
+
+## Package versions
+
+By default, when you `npm install` a [Buf module][modules], the BSR generates code from the most recent [reference](../overview.md#referencing-a-module) for the module. But you can also install a specific package version using npm's standard `@` syntax referencing an explicit version or a draft name.
+
+```terminal
+$ npm install @buf/protocolbuffers_js_acme_paymentapis@dev-branch
```
pkwarren, line 108: Looks like this content needs to be updated (using templates) or we at least need to move the previous version / syntax to another section of the doc.

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,224 @@
...
+`foo/bar.proto` | `some/other/path.ts` | ❌
+
+If you're a plugin author, be sure to heed this naming structure; otherwise, consumers of your APIs are likely to experience broken imports.
+
+[bsr]: /bsr/overview
+[bsr-plugins]: https://buf.build/plugins
+[bufbuild-plugins]: https://github.com/bufbuild/plugins
+[buf-npm]: https://buf.build/gen/npm/v1
+[deps]: /bsr/overview#dependencies
+[grpc-web]: https://buf.build/grpc/templates/web
```
pkwarren, line 204: This doesn't appear to be used?

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,224 @@
...
+[javascript]: https://javascript.com
+[labels]: /bsr/remote-generation/plugin-example#3-prepare-the-dockerfile
```
pkwarren, line 206: Unused.

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,224 @@
...
+
+If you're a plugin author, be sure to heed this naming structure; otherwise, consumers of your APIs are likely to experience broken imports.
+
+[bsr]: /bsr/overview
+[bsr-plugins]: https://buf.build/plugins
+[bufbuild-plugins]: https://github.com/bufbuild/plugins
+[buf-npm]: https://buf.build/gen/npm/v1
+[deps]: /bsr/overview#dependencies
+[grpc-web]: https://buf.build/grpc/templates/web
+[javascript]: https://javascript.com
```
pkwarren, line 205: Do we maybe want to point them to MDN https://developer.mozilla.org/en-US/docs/Web/JavaScript, ECMA script, or Wikipedia instead? This looks like a training site.

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,224 @@
...
+[labels]: /bsr/remote-generation/plugin-example#3-prepare-the-dockerfile
+[modules]: /bsr/overview#modules
+[npm]: https://npmjs.org
+[npm-config]: https://docs.npmjs.com/cli/v8/commands/npm-config#set
+[npmrc]: https://docs.npmjs.com/cli/v8/configuring-npm/npmrc
+[pb-go]: https://buf.build/protocolbuffers/templates/go
```
pkwarren, line 211: Unused.

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,224 @@
...
+[pb-js]: https://buf.build/protocolbuffers/templates/js
+[protoc]: https://github.com/protocolbuffers/protobuf
+[pnpm]: https://pnpm.io
+[ranges]: https://docs.npmjs.com/cli/v6/using-npm/semver#ranges
```
pkwarren, line 215: We can probably update this link to https://docs.npmjs.com/about-semantic-versioning.

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,243 @@
...
+$ npm install @buf/bufbuild_eliza.library_connect-web@0.2.1-20220706172350-dbde79169a01.3
+```
+
+## protoplugin support
+
+All plugins based on [`@bufbuild/protoplugin`][protoplugin] support the "rewrite_imports" option, which makes it possible to serve the output of a BSR module and a plugin in an individual package.
+
+This is desirable when plugins depend on code generated by other plugins to avoid duplicating base types.
+
+Example, the [`connect-web`][connect-web] plugin depends on base types generated by the [`protobuf-es`][protobuf-es] plugin. Since both plugins support the "rewrite_imports" option the BSR npm registry can direct the package manager to fetch dependent plugin code as a separate package.
```
timostamm, line 169: What do you think about linking to the plugins [@bufbuild/protoc-gen-connect-web](https://www.npmjs.com/package/@bufbuild/protoc-gen-connect-web) and [@bufbuild/protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es) on npmjs.com? 

It's how we refer to the plugins in various README files, with the rationale that the linked page gives an introduction, shows installation instructions and usage, and links to the source, so it's well suited for a canonical URL.

timostamm (commented):
The section on protoplugin support is spoton 👍 

I wonder if the section _Import rewrites_ rewrites below might confuse the reader. Maybe it makes sense to merge it with the section _protoplugin support_, for example with "Plugins that are not based on @bufbuild/protoplugin..."?

mfridman:
> The section on protoplugin support is spoton 👍
> 
> I wonder if the section _Import rewrites_ rewrites below might confuse the reader. Maybe it makes sense to merge it with the section _protoplugin support_, for example with "Plugins that are not based on @bufbuild/protoplugin..."?

Yes, great suggestion. I'll tweak the sections / wording (and also updated the example as suggested in the [other comment](https://github.com/bufbuild/docs.buf.build/pull/197#discussion_r972359136))

Originally I was thinking to remove the post-generation import rewrite section, but this might be valuable for users to understand.

mfridman:
Okay, took another stab at the import rewrites section in https://github.com/bufbuild/docs.buf.build/pull/197/commits/edb606576d6a75ad803b0f6fff9481c69fd7b7f8

I hope there is enough context in the first paragraph (Import path rewrites) and if a user is interested they can read the 2 sections separately:

- protoplugin support
- post generation import rewrites

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,243 @@
...
+### Import rewrites
+
+If the module you request has [dependencies][deps], the npm registry rewrites any relative import paths so that they point to the package with a full package name. Here's an example rewrite:
+
+```javascript
+// generated import
+require("../../google/storage/v1/storage_pb.js");
+
+// replacement
+require("@buf/grpc_web_googleapis_googleapis/google/storage/v1/storage_pb.js");
```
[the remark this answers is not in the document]
mfridman: Updated to use the acme/petapis as the example and fixed the naming.

We're planning to leave the existing section with a blurb linking to the new docs. We won't be doing a hard cutover but instead a gradual sunset of the existing npm registry v0.

Do you think we should have a short migration section? It might be useful.

timostamm (commented):
The import rewriting section looks good to me. But I'm curious if it makes sense to someone who hasn't been involved with it. @smaye81, what do you think about it?

> Do you think we should have a short migration section?

We should have migration docs _somewhere_. We will be recommending users to switch to curated plugins ([see PR](https://github.com/bufbuild/core/pull/6056)), so we should make it as easy as possible for them to migrate. 

For users of grpc/templates/web, protocolbuffers/templates/js, and bufbuild/templates/es, it means switching their dependency from the old package name to the new one. It should be a simple `npm remove` and `npm install`, and a search & replace for the package name in their application code.

For users of bufbuild/templates/connect-web, there is an additional step: Because the base types are a transitive dependency, all imports for base types in their application code need to be replaced. This is a manual change, because what was previously a single package name is now two package names ([PR in core](https://github.com/bufbuild/core/pull/5887) as an example).

Maybe Steve can chime in here too, he definitely knows his way around the connect-web generated code.

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,267 @@
+---
+id: npm-registry
+title: Npm registry
```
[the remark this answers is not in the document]
mfridman: Let's go with `NPM registry` 👍

smaye81:
> The import rewriting section looks good to me. But I'm curious if it makes sense to someone who hasn't been involved with it. @smaye81, what do you think about it?
> 
> > Do you think we should have a short migration section?
> 
> We should have migration docs _somewhere_. We will be recommending users to switch to curated plugins ([see PR](https://github.com/bufbuild/core/pull/6056)), so we should make it as easy as possible for them to migrate.
> 
> For users of grpc/templates/web, protocolbuffers/templates/js, and bufbuild/templates/es, it means switching their dependency from the old package name to the new one. It should be a simple `npm remove` and `npm install`, and a search & replace for the package name in their application code.
> 
> For users of bufbuild/templates/connect-web, there is an additional step: Because the base types are a transitive dependency, all imports for base types in their application code need to be replaced. This is a manual change, because what was previously a single package name is now two package names ([PR in core](https://github.com/bufbuild/core/pull/5887) as an example).
> 
> Maybe Steve can chime in here too, he definitely knows his way around the connect-web generated code.

Mike sent me the text earlier today.  I mentioned that it makes sense, it just feels a bit complicated.  However, that could very well be a function of how difficult the import rewrite logic is.  I mentioned that regardless this is a good starting point and it will be nice to have docs to point people to if/when they ask questions.  We can always refine the docs if we realize they are not clear as people being to consume them.

docs/bsr/remote-generation/migration-guide.md
```diff
@@ -0,0 +1,141 @@
...
+
+The naming convention has changed because templates have been removed in favor of plugins. The new format is:
+
+`{moduleOwner}_{moduleName}.{pluginOwner}_{pluginName}`
+
+Note the dot (`.`) delimiter is used to break up the module and the plugin components.
+
+This means you'll need to do 2 things: 
+
+1. `npm remove` the old package and `npm install` the new package
```
smaye81, line 85: I think you want `npm uninstall` here right?  I don't think `npm remove` is a command.

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,261 @@
...
+  ]}
+/>
+
+You can use an existing auth token or generate a new one. To create a new one, log into the [BSR][bsr], navigate to your [user settings][settings] page, and click **Create Token**.
+
+## Available plugins {#plugins}
+
+For a full list of supported plugins navigate to the [BSR plugins page][bsr-plugins] and search for JavaScript/TypeScript.
+
+To learn more about how these plugins are packaged and distributed check out the [bufbuild/plugins repository][bufbuild-plugins]. If you find a useful plugin that should be added please file an issue!
```
smaye81, line 62: Same as above re: linking to the issue page.

docs/bsr/remote-generation/npm-registry.md
```diff
@@ -0,0 +1,261 @@
...
+} />
+
+In this example, `@buf/bufbuild_eliza.library_connect-web`, the BSR generates code for the [`bufbuild_eliza`](https://buf.build/bufbuild/eliza) module against the [`library_connect-web`](https://buf.build/library/connect-web) plugin.
+
+## Package versions
+
+By default, when you `npm install` a [Buf module][modules], the BSR generates code from the most recent [reference](../overview.md#referencing-a-module) for the module. But you can also install a specific package version using npm's standard `@` syntax referencing an explicit version or a draft name.
+
+```terminal
+$ npm install @buf/bufbuild_eliza.library_connect-web
```
smaye81, line 107: Should this have a version on the end of it?  i.e. `npm install @buf/bufbuild_eliza.library_connect-web@0.1.1`?

mfridman:
Okay, I added a migration guide. Tried to keep it minimal.

docs/bsr/remote-generation/migration-guide.md
mfridman: `npm` .. where every command has a half-dozen aliases. `npm install`:

aliases: remove, rm, r, un, unlink

But to your point, I'll change this to `uninstall` because that's the more common one.

docs/bsr/remote-generation/npm-registry.md
mfridman: I was hoping to keep this short, hence the lack of a version. For most users the default is `latest` based on their config tag.

> If you ask npm to install a package and don't tell it a specific version, then it will install the specified tag.

[Reference](https://docs.npmjs.com/cli/v8/using-npm/config#tag)

srikrsna (approved):

PR #198, opened 2022-09-15
add draft related docs

cyinma:
fixes [BSR-469](https://linear.app/bufbuild/issue/BSR-469)


docs/bsr/overview.md
```diff
@@ -74,7 +74,7 @@ deps:
   - buf.build/acme/units
 ```
 
-Although we **do not recommend** it, in some situations you may need to pin a module to a specific version. Ideally, authors keeps modules backwards-compatible and avoid breaking changes so you can *always* rely on the latest version.
+Although we **do not recommend** it, in some situations you may need to pin a module to a specific version, you can do so by specifying tag, commit, or draft in your `deps` after the `:` delimiter. Ideally, authors keeps modules backwards-compatible and avoid breaking changes so you can *always* rely on the latest version.
```
mfridman, line 77: Is this correct, i.e., can we specify a dependency by its draft name?

```
deps:
  - buf.build/acme/paymentapis:dev-branch
```

Iirc this will error:

> ... a module with a dependency pinned to a draft commit is not allowed

docs/bsr/overview.md
```diff
@@ -110,12 +110,14 @@ The `buf` CLI automatically resolves the module(s) specified in the `deps` list.
...
 
-A reference is a way to refer to a single version of the repository. While a reference always _resolves_ to a single snapshot of the repository, it can be either a commit or a tag.
+A reference is a way to refer to a single version of the repository. While a reference always _resolves_ to a single snapshot of the repository, it can be either a commit, a tag, or a draft.
 
 **Commit**: Every push of new content to a repository is associated with a commit that identifies that change in the schema. The commit is created after a successful push. This means that unlike Git, the commit only exists on the BSR repository and not locally.
 
 **Tag**: A reference to a single commit but with a human readable name, similar to a Git tag. It is useful for identifying commonly referenced commits&mdash;like a release.
 
+**Draft**: A temporary commit in development workflow with a human readable name, similar to a Git feature branch but without history. It is useful for developing on the module and want to commit outside the main branch. Can be replaced by another commit with the same draft name and is deletable.
```
mfridman, line 119: Maybe worth adding a note that drafts cannot be used for dependency resolution?

docs/tour/add-a-dependency.md
```diff
@@ -125,7 +125,7 @@ the imports required to build your module. **You don't have to manually copy `.p
 
 ## 9.3 Pin Your Dependencies {#pin-your-dependencies}
 
-You can pin to a specific tag or commit by specifying it in your `deps` after the `:` delimiter. For example,
+You can pin to a specific tag, commit, or draft by specifying it in your `deps` after the `:` delimiter. For example,
```
mfridman, line 128: Similar to first comment, is this true? This seems to conflict with "Push a module" section..

> ... and the module does not have any dependency that is pinned to a draft commit

docs/bsr/overview.md
cyinma: yeh we changed the behaviour to allow using draft as dependency, but disallow it on push

mfridman: I see, so drafts work everywhere _except_ for the push command?

docs/bsr/overview.md
mfridman: Ye, I think we need some blurb (probably in this section) that explicitly says modules with dependencies referencing a draft cannot be pushed to the BSR. 

Tbh I came away a bit confused because this wasn't immediately obvious to me (unless I missed it).

docs/bsr/overview.md
cyinma: yeh basically it just works like all other references except you cannot push a module with a draft dependency

docs/bsr/overview.md
```diff
@@ -110,12 +110,14 @@ The `buf` CLI automatically resolves the module(s) specified in the `deps` list.
...
+A reference is a way to refer to a single version of the repository. While a reference always _resolves_ to a single snapshot of the repository, it can be either a commit, a tag, or a draft.
 
 **Commit**: Every push of new content to a repository is associated with a commit that identifies that change in the schema. The commit is created after a successful push. This means that unlike Git, the commit only exists on the BSR repository and not locally.
 
 **Tag**: A reference to a single commit but with a human readable name, similar to a Git tag. It is useful for identifying commonly referenced commits&mdash;like a release.
 
+**Draft**: A temporary commit in development workflow with a human readable name, similar to a Git feature branch but without history. It is useful for developing on the module and want to commit outside the main branch. Can be replaced by another commit with the same draft name and is deletable. When it is used as a dependency in `buf.yaml` of a module, the module cannot be pushed until you updated to a non-draft commit of the dependency.
```
mfridman, line 119:
```suggestion
**Draft**: A temporary commit in development workflow with a human readable name, similar to a Git feature branch but without history. It is useful for iterating on a module while keeping those changes outside the main branch. Can be replaced by another commit with the same draft name and is deletable. When it is used as a dependency in `buf.yaml` of a module, the module cannot be pushed until you update to a non-draft commit of the dependency.
```

docs/tour/add-a-dependency.md
```diff
@@ -171,3 +171,5 @@ $ buf push
 ---
 b2917eb692064beb92ad1e38dba6c25e
 ```
+
+> If the dependencies is pinned to a draft commit, pushing the module will not be allowed.
```
mfridman, line 175:
```suggestion
> If one or more dependencies are pinned to a draft commit, pushing the module will not be allowed.
```

mfridman (approved):
Small suggestions, otherwise lgtm.

mfridman:
Thanks for putting this together, let's merge this into the feature branch and keep iterating on it and gathering feedback.

PR #208, opened 2022-09-30
Fix prettierrc and lint files

renahlee:
This PR:

- Changes pages with swizzled components to `.mdx`
- Re-enables linting for `/docs` so that we have consistent whitespace (removes deprecated field `jsxBracketSameLine`;  minimizes whitespace churn for oncoming copy edits for `repository` ➡️  `module` / `commit` ➡️  `snapshot` change)
- Changes `printWidth` from `100` to `80` (more standard)


docs/bsr/remote-generation/go.md
```diff
@@ -10,67 +10,64 @@ import Breaking from "@site/src/components/Breaking";
...
-/>
+<Syntax title="Generated Go module path syntax" examples={[
+"go.buf.build/grpc/go/bufbuild/eliza" ]} segments={[ {"label": "go.buf.build",
+"kind": "constant"}, {"separator": "/"}, {"label": "templateOwner", "kind":
+"variable", href: "/bsr/remote-generation/overview#templates"}, {"separator":
+"/"}, {"label": "templateName", "kind": "variable", href:
+"/bsr/remote-generation/overview#templates"}, {"separator": "/"}, {"label":
+"moduleOwner", "kind": "variable", href: "/bsr/overview#modules"}, {"separator":
+"/"}, {"label": "moduleName", "kind": "variable", href:
+"/bsr/overview#modules"}, ]} />
```
renahlee, line 49: 👀

.prettierignore
```diff
@@ -1,4 +1,10 @@
 .docusaurus
 build
-docs
 drafts
+docs/bsr/remote-generation/hosted-plugins.md
+docs/bsr/remote-generation/js.md
+docs/bsr/remote-generation/go.md
```
timostamm, line 6: I agree with the changes, but have a couple of questions:

It looks like these pages use JSX. Is prettier breaking them, or is there another reason we have to prettierignore them?

I wonder if it is a good idea to format some md files, but not others. It is not immediately clear under what circumstances the formatter can be used. What happens if someone wants to use JSX, after seeing it used on another page?

I also think that using a formatter on the md files will cause issues when it is not enforced by CI. Some contributors will use prettier, others will not. This means (sooner or later) a contributor aware of the formatter will have unexpected diffs in unrelated files.

renahlee: > It looks like these pages use JSX. Is prettier breaking them, or is there another reason we have to prettierignore them?

Unfortunately prettier is breaking these files and causes build failures but there's a chance I missed some kind of formatting trick or other prettier rule where I can avoid ignoring these files. Let me know if you know of anything :)

> I wonder if it is a good idea to format some md files, but not others. It is not immediately clear under what circumstances the formatter can be used. What happens if someone wants to use JSX, after seeing it used on another page?

I thought about this as well -- a not great answer I have is to lint by default and manually fix the usage of swizzled components if there isn't a better solution out there.

> I also think that using a formatter on the md files will cause issues when it is not enforced by CI. Some contributors will use prettier, others will not. This means (sooner or later) a contributor aware of the formatter will have unexpected diffs in unrelated files.

This is unfortunately already the case -- there were files where folks fell back on their prettier defaults (hence why some files were wrapped at `80`, others `100`, and some not at all).

renahlee: For posterity, here's an example of what happens:

```
import Syntax from "@site/src/components/Syntax";

<Syntax
  title="Lint error syntax"
  examples={[
    "pet/v1/pet.proto:47:9:Service name \"PetStore\" should be suffixed with \"Service\"."
  ]}
  segments={[
    {label: "file", kind: "variable"},
    {separator: ":"},
    {label: "line", kind: "variable"},
    {separator: ":"},
    {label: "column", kind: "variable"},
    {separator: ":"},
    {label: "message", kind: "variable"}
  ]}
/>
```

becomes:

```
import Syntax from "@site/src/components/Syntax";

<Syntax title="Lint error syntax" examples={[ "pet/v1/pet.proto:47:9:Service
name \"PetStore\" should be suffixed with \"Service\"." ]} segments={[ {label:
"file", kind: "variable"}, {separator: ":"}, {label: "line", kind: "variable"},
{separator: ":"}, {label: "column", kind: "variable"}, {separator: ":"}, {label:
"message", kind: "variable"} ]} />
```

timostamm: It looks like we can mitigate both problems (formatter breaking some md files, contributors forgetting to format) by adding CI that formats, and fails if there is a diff. 

Thinking this through, there should be a make target to format, a note in the readme about the need to prettierignore md files with JSX, and CI that calls `make format`, then checks for a diff (see [here](https://github.com/bufbuild/connect-go/blob/062122ca56ef77e244f50375db17837dab748dfc/Makefile#L76-#L78) for a nice example).

This doesn't sound _great_, but also not terrible. 

I only have one idea for prettier and JSX: To give the md files with JSX a `.mdx` extension. Prettier [appears to support it](https://prettier.io/blog/2021/11/25/2.5.0.html), but I have never tried it. If that works, we might not _need_ CI after all.

timostamm: While we do use mdx in the connect docs, we don't run prettier on them either. So the main question whether prettier support for mdx is mature enough is still open.

renahlee: @timostamm your suggestion of using `.mdx` seems to be the right solution forward. I'll update this PR today but need help testing as my local docs env is broken.

docs/bsr/remote-generation/hosted-plugins.mdx
```diff
@@ -87,6 +121,12 @@ import TabItem from '@theme/TabItem';
   ]}>
   <TabItem value="go">
 
+```yaml title=buf.gen.yaml
+version: v1
+  - remote: buf.build/grpc/plugins/python:v1.41.1-1
+    out: gen/python
+```
+
```
timostamm, line 129: This doesn't look right.

docs/bsr/remote-generation/overview.mdx
```diff
@@ -0,0 +1,299 @@
...
+definitions. Examples of Protobuf plugins include
+[`protoc-gen-go`][protoc-gen-go] and [`protoc-gen-python`][protoc-gen-python].
+
+:::info
+
+Creating your own plugins See the documentation on
+[authoring plugins](./plugin-example.md) to see how you can create and upload
+your own plugins to use as part of code generation.
+
+:::
```
timostamm, line 73: "Creating your own plugins" was the admonition title previously:

```
:::info Creating your own plugins
```

This seems to a known issue with [docusaurus and prettier](https://docusaurus.io/docs/markdown-features/admonitions#usage-with-prettier), but there is a workaround. Can you check other admonitions?

docs/build/usage.md
```diff
@@ -221,7 +221,7 @@ $ buf build --error-format=json
 By default, `buf build` outputs its result to `/dev/null`. In this case, it's common to use
 `buf build` as a validation step, analogous to checking if the input compiles.
 
-`buf build` also supports outputting [`FileDescriptorSet`s](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto)
+`buf build` also supports outputting [`FileDescriptorSets`](https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/descriptor.proto)
```
timostamm, line 224: `FileDescriptorSet` is a well-known type, but `FileDescriptorSets` is not 🤔  

`Image` is also a type, but we have a simple link to "Images". I think it would be fair to make "FileDescriptorSets" a simple link as well:

Screen Shot 2022-10-06 at 13 29 21

timostamm, line 224: `FileDescriptorSet` is a well-known type, but `FileDescriptorSets` is not 🤔  

`Image` is also a type, but we have a simple link to "Images". I think it would be fair to make "FileDescriptorSets" a simple link as well:

Screen Shot 2022-10-06 at 13 29 21

docs/build/usage.md
```diff
@@ -317,7 +317,9 @@ dependent descriptors are included in the build:
   - Request and response types referenced in service methods
   - Any custom options for the services, its methods, and the file in which the service is defined
 
-:::success Supplying multiple types
+:::success
+
+Supplying multiple types
```
timostamm, line 322: Broken admonition title.

timostamm (commented):
Interesting to see this normalizing quotes and yaml indentation at quite a few places. Great stuff! I went through about a third and noticed that some admonition titles broke. Could you give this a closer look?

docs/bsr/remote-generation/overview.mdx
renahlee: Ah this is my bad -- yes, prettier does indeed break admonitions but the workaround is to add the extra line. I think the step I need to do is move the title back to the admonition type level.

docs/build/usage.md
renahlee: Ah thank you for clarifying! This is my bad -- I thought it was a typo ([`FileDescriptorSet`s]) -- the correct change here is actually likely to move the `s` out of the link entirely.

timostamm: Good call, this might work even better.

timostamm (approved):

PR #210, opened 2022-10-04
Repository clarification

joshcarp:


docs/bsr/overview.md
```diff
@@ -40,7 +40,7 @@ import Syntax from "@site/src/components/Syntax";
 
 - **Remote**: The DNS name for the server hosting the BSR. This is always `buf.build`.
 - **Owner**: An entity that is either a user or organization within the BSR ecosystem.
-- **Repository**: Stores a single module and all versions of that module.
+- **Repository**: Stores all the different versions of the same module.
```
renahlee, line 43: Nit suggestion: stores all versions of a module

joshcarp: I do want to strongly emphasize that it's *different* versions of the *same* module. `Stores all versions of a module` fails to imply that a repository only stores one module

docs/bsr/overview.md
```diff
@@ -108,9 +108,9 @@ The `buf` CLI automatically resolves the module(s) specified in the `deps` list.
 
 ## Referencing a module
 
-Each module on the BSR exists as a snapshot, and contains a unique reference associated with every change.
+Each module in a BSR repository exists as a snapshot, and contains a unique reference associated with every change.
```
renahlee, line 111: Suggestion: A module exists as snapshots in a BSR repository. A snapshot can include any number of changes, where each change is identifiable by a unique reference.

docs/bsr/overview.md
```diff
@@ -108,9 +108,9 @@ The `buf` CLI automatically resolves the module(s) specified in the `deps` list.
...
 
-A reference is a way to refer to a single version of the repository. While a reference always _resolves_ to a single snapshot of the repository, it can be either a commit or a tag.
+A reference is a way to refer to a single version of the module. While a reference always _resolves_ to a single version of the module, it can be either a commit or a tag.
```
renahlee, line 113: (If the above suggestion is accepted), remove `A reference is a way to refer to a single version of the module`, then:

```
While a reference always _resolves_ to a single snapshot, it can be either a commit or a tag.
```

docs/bsr/overview.md
renahlee: Hmmm perhaps, `stores all version of a single module` would address that?

docs/bsr/usage.md
```diff
@@ -16,7 +16,7 @@ Follow the prompts to enter your username and password (API Token). This adds an
 The only requirement to push to the BSR is to have a module that [builds](../build/usage.md)
 successfully, meaning that you should be able to run `buf build` successfully on your module.
 
-Before you can push a module into the BSR there needs to exist a repository, owned by either a user or an organization, the user has access to. A repository can be created either through the UI or from the command line with `buf`.
+Before you can push a module into the BSR a repository must exist, owned by either a user or an organization, the user has access to. A repository can be created either through the UI or from the command line with `buf`.
```
renahlee, line 19: Minor grammar change (removal of the comma after `organization`): ...owned by either a user or an organization the user has access to

renahlee (approved):

docs/bsr/overview.md
renahlee, line 43:
```suggestion
- **Repository**: Stores all versions of a single module
```

docs/bsr/overview.md
renahlee, line 111:
```suggestion
A module exists as snapshots in a BSR repository. A snapshot can include any number of changes, where each change is identifiable by a unique reference.
```

docs/bsr/overview.md
renahlee, line 113:
```suggestion
While a reference _always_ resolves to a single snapshot, it can be either a commit or a tag.
```

docs/bsr/usage.md
renahlee, line 19:
```suggestion
Before you can push a module into the BSR, a repository must exist and the user must own or have access to that repository. A repository can be created either through the UI or from the command line with `buf`.
```

docs/bsr/overview.md
joshcarp: imo we should ditch the use of the word "snapshot" and just use "version"

renahlee: I agree with this but want to be consistent with the above definition. If we remove usage of `snapshot` above, then I'm down for `version` here.

renahlee: I agree with this. There was already mixed usage in docs where folks referred to `commit` as `version`, which means we're predisposed to thinking of it as a `version`.

renahlee, line 111:
```suggestion
A module has different versions. Each version includes any number of changes, and each change is identifiable by a unique commit or tag. The collective set of module versions is housed within a repository.
```

docs/bsr/overview.md
renahlee, line 113:
```suggestion
```

bufdev (approved):

SyrieBianco (commented):
Looks good to me! 

For the paper trail, related to the [BSR Concept Renaming](https://www.notion.so/bufbuild/BSR-concept-renaming-cd261e6412d5497189892677d31d2dc6) eng doc and the now cancelled [Repository -> Module, Commit -> Snapshot](https://www.notion.so/bufbuild/Repository-Commit-Module-Snapshot-Rename-34626bade3224351bdfe732ab525948b) project

PR #215, opened 2022-10-07
Remove additional h1 from Buf Studio docs

stefanvanburen:
Just another small tweak - we currently have two `h1`s here, right next to each other, which is a bit odd. In line with the other docs, use the shorter one, since we also use "interactive web UI for gRPC and protobuf" in the first paragraph.

Currently:

image


renahlee (approved):

renahlee:
Thanks so much for all the docs cleanup ❤️

PR #221, opened 2022-10-14
Migrated to a custom stars plugin over env var

paul-sachs:
- created a tiny little docusaurus plugin so we can access the star count instead of our jank current setup (bonus, is reusable)
- added node-fetch so we can use fetch in node 16


renahlee (approved):
Thank you 🚀🚀🚀

PR #223, opened 2022-10-18
generate-code: clarify `buf.gen.yaml` location

jchadwick-buf:
In the generate-code part of the tour, it is somewhat ambiguous where the `buf.gen.yaml` file should be placed. This change tweaks the verbiage to spell it out more clearly so that later commands work as-expected.

(It's a minor issue, but admittedly I put the `buf.gen.yaml` next to the `buf.yaml` initially just by instinct, so I'm sure a non-zero amount of other people will do the same.)


renahlee:
CC: @doriable for visibility

renahlee (approved):

PR #225, opened 2022-10-24
Fix links to descriptor.proto from Managed Mode docs

doriable:
This fixes the links to `descriptor.proto` in managed mode docs
to be in the correct line numbers.


doriable:
> looks good - I guess we don't permalink (i.e. https://github.com/protocolbuffers/protobuf/blob/fcd3b9a85ef36e46643dc30176cea1a7ad62e02b/src/google/protobuf/descriptor.proto#L416) to not fall too far out of date?

Yes, I believe that's the case, although I will say, it's a bit of a pain. I just updated this for a few other links on the page (figured I might as well while I was here). :)

stefanvanburen (approved):

PR #236, opened 2022-10-25
Fix Studio mermaid diagram

stefanvanburen:
Currently these are just rendering as text:

image

This just reverts them to https://github.com/bufbuild/docs.buf.build/blob/c2ee7631711a7e87838a880ef7dd2cd57ab2076c/docs/bsr/studio.md.

I think the Prettier updates in #208 might have been the culprit - I'm not sure how we're running prettier (could be nice to include in `make lint`?), but do I need to do something here to get it to ignore these lines? cc @renahlee


renahlee:
> do I need to do something here to get it to ignore these lines? cc @renahlee

You can add them to the `.prettierignore` -- just add a comment linking to [FE-525](https://linear.app/bufbuild/issue/FE-525/mermaid-diagrams-dont-render-correctly-if-file-is-mdx). Thanks for getting this fixed up! ❤️

renahlee (approved):

stefanvanburen:
> > do I need to do something here to get it to ignore these lines? cc @renahlee
> 
> You can add them to the `.prettierignore` -- just add a comment linking to [FE-525](https://linear.app/bufbuild/issue/FE-525/mermaid-diagrams-dont-render-correctly-if-file-is-mdx). Thanks for getting this fixed up! ❤️

I could do that, but it would be nice to have prettier work on it. Maybe it should be an `.mdx` file given that [it's importing components](https://github.com/bufbuild/docs.buf.build/blob/ea0d080c37a523f67fe47b5c4b47d30ef9ead550/docs/bsr/studio.md?plain=1#L6-L7)?

stefanvanburen:
Chatted offline, Mermaid components don't render quite right if they're in a .mdx currently, so for now I'll merge this as is and we can investigate that separately!

PR #252, opened 2022-11-02
Clarify docs on ref

robbertvanginkel:
As many git providers don't allow fetching individual shas, ref is unfortunately limited to checking out references in the fetched history. Clarify this in the doc and add notes on how to change the initial clone parameters to checkout refs outside the default clone scope.


mfridman (approved):
Good update!

PR #264, opened 2022-11-23
Update Generated documentation page

seankimdev:
- Adds a section on the rendered Protobuf options in the generated documentation
- Updates the screenshots from the `Generated documentation` page to reflect the current UI.
- Change the styling for figure caption texts.


mfridman:
Looking great, any chance we can update the screenshots to not reference demolab (we want to delete this eventually).

Maybe we can use petapis or eliza instead?

seankimdev:
> Maybe we can use petapis or eliza instead?

Yeah of course! Thank you for the info.

doriable (approved):
Looks good to me! 🙏🏼  I think we should change the `gen_docs-3.png` screenshot to use Eliza, but stamping first, since that does not require re-review. Thanks for updating this!

mfridman (approved):

PR #280, opened 2022-12-06
include information on removing revision

pkwarren:
When migrating from the alpha, users should remove any revisions specified for the plugin.


timostamm (approved):
Looks good to me!

mfridman:
Do we want to advertise that it _is_ possible to pin to a specific plugin + version + revision? (not sure if it's in the docs)

Or do we want folks to always use latest for a given version?

pkwarren:
> Do we want to advertise that it _is_ possible to pin to a specific plugin + version + revision in `buf.gen.yaml`? (not sure if it's in the docs)

It is in the linked docs for buf.gen.yaml. The main issue with recommending keeping it is that the revisions of old plugins don't necessarily match the revisions of newer plugins. I think in most cases it should be omitted.

PR #281, opened 2022-12-07
Fix makefile

elliotmjackson:
makefile release step referenced an incorrect file path, failing the [run](https://github.com/bufbuild/docs.buf.build/actions/runs/3642158951/jobs/6148938669).


joshcarp (approved):
LGTM

PR #291, opened 2022-12-12
fix buf.gen.yaml examples

### LINKED ISSUE #290: Buf-Tour: Inconsistency for relative vs. absolute paths in code generation with remote plugins step
There is an inconsistency between the `buf.gen.yaml` config and the output in the remote plugin step (https://github.com/bufbuild/docs.buf.build/blob/316ea8d28e3366b2949d817bca4e0fcad791f2b8/docs/tour/generate-go-code.md#106-edit-your-bufgenyaml-edit-your-bufgenyaml-remote)

When applying the given diff, the `source_relative` option is also removed, resulting in fully qualified package/folder structures when performing the (remote) code generation.

In addition to that, there is a minor inconsistency: The change from `gen/proto/go` to `gen/go` happens in the same step without being reflected in the diff itself although it changed since step 10.3.

Happy to support fixing these docs. Which of the alternatives (keeping option vs. documenting full tree structure) would be preferred?

pkwarren:
Fix several issues with examples/diffs in remote plugins generation.

Fixes #290.


elliotmjackson (approved):

docs/bsr/remote-plugins/overview.mdx
```diff
@@ -31,9 +31,11 @@ plugins:
   # Use protoc-gen-go at v1.28.1
   - plugin: buf.build/protocolbuffers/go:v1.28.1
     out: gen/go
+    opt: paths=source_relative
   # Use the latest version of protoc-gen-go-grpc
   - plugin: buf.build/grpc/go
     out: gen/go
+    opt: paths=source_relative
```
pkwarren, line 38: We should always leave these specified. I think we may have removed when plugins had `default_opts`.

docs/generate/usage.mdx
```diff
@@ -81,7 +81,9 @@ plugins:
     opt: paths=source_relative
   - plugin: go-grpc
     out: gen/go
-    opt: paths=source_relative
+    opt:
+      - paths=source_relative
+      - require_unimplemented_servers=false
```
pkwarren, line 86: Made the first example match the second diff.

docs/generate/usage.mdx
```diff
@@ -158,19 +160,19 @@ you want to use. We can now modify the `buf.gen.yaml` for
 [grpc](https://buf.build/grpc/go), remote plugins so `protoc-gen-go` and
 `protoc-gen-go-grpc` do not need to be in your `$PATH`:
 
-```diff title="buf.gen.yaml" {3,4,6,7,8,10,11,12}
+```diff title="buf.gen.yaml" {3,4,7,8}
```
pkwarren, line 163: Best to look at this in the preview, but now the only diffs are to the `plugin:` lines.

docs/generate/usage.mdx
```diff
@@ -234,8 +236,8 @@ $ buf generate --template buf.gen.java.yaml
 You could also specify those different templates as JSON:
 
 ```terminal
-$ buf generate --template '{"version":"v1","plugins":[{"name":"go","out":"gen/go"}]}'
-$ buf generate --template '{"version":"v1","plugins":[{"name":"java","out":"gen/java"}]}'
+$ buf generate --template '{"version":"v1","plugins":[{"plugin":"go","out":"gen/go"}]}'
+$ buf generate --template '{"version":"v1","plugins":[{"plugin":"java","out":"gen/java"}]}'
```
pkwarren, line 240: Missed these as part of `name`/`remote` -> `plugin` updates.

PR #292, opened 2022-12-13
BSR-1049 Add wkt to faq.md

oliversun09:
Add _Depending on well-known types_ section to `faq.md`.

This section was pointed to by [beta/wellknowntypes](https://buf.build/beta/wellknowntypes) but did not exist.


docs/faq.md
```diff
@@ -138,3 +138,19 @@ $ protoc -I . \
 ```
 
 We apologize for any inconvenience this warning may have caused.
+
+## Depending on well-known types {#why-cant-i-import-bufbuildbetawellknowntypes-as-a-dependency-in-my-module}
+
+You may have seen this error:
+
+```
+Failure: repository beta/wellknowntypes cannot be resolved as a dependency, see https://buf.build/beta/wellknowntypes for more information
```
stefanvanburen, line 147: do you have an example of where a user would see this message? just a `buf` command using a buf.yaml with `buf.build/beta/wellknowntypes` in its `deps`?

stefanvanburen (approved):
looks good to me, just wondering where we would see that specific error message?

docs/faq.md
oliversun09: Added to the doc that it could happen when running `buf mod update`

stefanvanburen (approved):

PR #312, opened 2023-01-05
Fix slug concatenation bug

elliotmjackson:
If a user navigates to the remote packages overview page from the /plugins/n page of the BSR, they will be taken to `/bsr/remote-packages/overview/` NOTE the trailing `/`
image
when navigating to any of the child links, they will be appended to the end of the slug making the link `/bsr/remote-packages/overview/` + `(go|npm)` which does not exist.  
image
using an absolute path here resolves the bug


stefanvanburen (approved):
👍 fixes the issue

PR #360, opened 2023-02-03
Add npm install instructions

paul-sachs:
Add npm installation instructions. I was thinking we might want to also add a note about adding `node_modules` to the path in buf.gen.yaml. Does that make sense for a version of buf installed into a projects node_modules directory?


timostamm (commented):
We should not recommend global installation for several reasons, some are explained [here](https://medium.com/@maybekatz/introducing-npx-an-npm-package-runner-55f7d4bd282b).

There is also another reason we want the local installation: code generator plugins. We don't want to add a note about adding `node_modules` to the path, because it might not even exist with pnpm or yarn, and with workspaces, you might have to reference a parent `node_modules`. The [readme for protoc-gen-es](https://www.npmjs.com/package/@bufbuild/protoc-gen-es) shows a workaround for yarn, and it's not great DX.

Instead, we should recommend _local_ installation. That way, you can simply install buf and plugins:

```shell
npm i @bufbuild/buf @bufbuild/protoc-gen-es
```

And you just need the short plugin name (prefix `protoc-gen-` removed) in your `buf.gen.yaml`:

```yaml
version: v1
plugins:
  - name: es
    out: src/gen
```

To generate code with

```bash
npx buf generate
```

Because `npx` puts all npm "binaries" on the $PATH.

That's the DX we want for code gneration. That's why we needed buf on npm.

paul-sachs:
@timostamm yeah that makes sense. The `npx` goal doesn't quite work right now and I wasn't sure why. Investigating, it turns out that the package MUST be the same name as the binary which doesn't currently work right now since our base package is `@bufbuild/buf`. Guess that means i'll make an adjustment to publish it as `buf` instead. I'll make some adjustments over in the publishing repo.

paul-sachs:
Ah crap, https://www.npmjs.com/package/buf already exists...

paul-sachs:
So i think the cleanest we can get it right now is `npx --package @bufbuild/buf buf generate`

paul-sachs:
Ah, right, i guess that's why you mentioned [bufbuild](https://www.npmjs.com/package/bufbuild) in notion.

paul-sachs:
Alright, updated docs to be project focused.

timostamm:
Please try to run the example I gave, Paul. It works 😄

Note that you can also run `npx tsc` if you installed the `typescript` package. `ls node_module/.bin` to get a hint what's happening.

paul-sachs:
@timostamm oh damn! The docs around [npx](https://www.npmjs.com/package/npx) really don't make that clear at all! Good to know. I'll update the docs accordingly.

docs/installation.mdx
```diff
@@ -30,6 +30,23 @@ This installs:
...
+$ npm install @bufbuild/buf
+```
+
+This installs the binaries `buf`, `protoc-gen-buf-breaking`,
+`protoc-gen-buf-lint` to be used within your project.
+
+To execute `buf` from the command line, you can use `npx`:
+
+```terminal
+npx buf generate
```
timostamm, line 47: Would `npx buf` or `npx buf --version` be a better example for invoking through `npx`? If you have only just installed the CLI, you likely don't have a `buf.gen.yaml` yet.

timostamm (approved):
Not sure whether `npx buf generate` is the best example in this context, see comment below.   Maybe @elliotmjackson would like to chime in here, but either way, LGTM from me.

docs/installation.mdx
paul-sachs: Fair. I think switching it to just "--version" is fine by me.

PR #366, opened 2023-02-07
Update buf-gen-yaml.md

tonyli233:
Updated documentation as per the change in https://github.com/bufbuild/buf/pull/1792


docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -116,7 +132,21 @@ plugins:
     path: bin/proto/protoc-gen-foo
 ```
 
-This field only works with local plugins.
+The `path` is **required** when the `name` field is specified. The `path` is
```
jhump, line 135: I don't believe this is true. If no path is given, it will assume a binary named `protoc-gen-{name}` exists in the `PATH`.

docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -48,25 +48,40 @@ compiled representation of your module.
...
+executed plugin with the `plugin` field. You **cannot** set the [path](#path) 
+and [strategy](#strategy) option when the `remote` option is used.
+
+#### `name`
+
+The `name` field can be any user-specified name of the plugin. The [path](#path)
+option is **required** if the `name` option is specified. The path of the plugin
+will be executed to generate codes. 
+
+*Note*: Exactly one of `plugin`, `remote`, or `name` is required.
```
jhump, line 84: Docusaurus supports a better style of [callout](https://docusaurus.io/docs/markdown-features/admonitions):

```suggestion
:::note
Exactly one of `plugin`, `remote`, or `name` is required.
:::
```

But, having said that, are we sure we want to add documentation for the other fields? It was my understanding that we deprecated them: I thought you could do everything just with the `plugin` property, and the CLI would detect if it's a remote or local plugin.

jhump (commented):
I think @bufdev needs to cast an eye on this. I think the lack of descriptions for `remote` and `name` may have been intentional, but I'm not sure.

docs/configuration/v1/buf-gen-yaml.md
tonyli233: Yeah I just checked we don't have to specify `path`.

tonyli233:
> I think @bufdev needs to cast an eye on this. I think the lack of descriptions for `remote` and `name` may have been intentional, but I'm not sure.

I added the documentation for these fields since there were missing in the original docs. I'll wait for answer from @bufdev.

bufdev:
Yea why is there such a big diff? I deleted a lot of deprecated fields in https://github.com/bufbuild/docs.buf.build/pull/353. Don't overwrite these changes, just update for this new feature.

tonyli233:
> Yea why is there such a big diff? I deleted a lot of deprecated fields in #353. Don't overwrite these changes, just update for this new feature.

Just corrected this. I didn't realize those were deleted in previous PRs.

bufdev (approved):

PR #367, opened 2023-02-07
Update buf.gen.yaml for protoc_path

tonyli233:
Updated buf.gen.yaml for protoc_path field.


docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -201,6 +201,20 @@ $ protoc -I . $(find . -name '*.proto')
 This is needed for certain plugins that expect all files to be given at once.
 It is also used for remote plugin generation to improve code generation performance.
 
+#### `protoc_path`
+The `protoc_path` of a plugin is optional, and overrides what `protoc` points to in
```
bufdev, line 205: We need commentary for what plugins this affects, and why protoc is invoked, this is a speciality option only for the built in plugins. One of us may need to help draft this

docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -201,6 +201,20 @@ $ protoc -I . $(find . -name '*.proto')
 This is needed for certain plugins that expect all files to be given at once.
 It is also used for remote plugin generation to improve code generation performance.
 
+#### `protoc_path`
+The `protoc_path` of a plugin is optional, and overrides what `protoc` points to in 
+`PATH`. For example, you can run an old version of `protoc` instead of the `protoc`
+installed in `PATH` by:
```
jhump, line 207:
```suggestion
The `protoc_path` of a plugin is optional and only applies to the code generators
that are _built-in_ to `protoc`. Normally, a plugin is a separate executable whose
binary name is like `protoc-gen-<name>`. But for a handful of plugins, the executable
used is `protoc` itself. The following plugins result in invoking `protoc` instead of
a dedicated plugin binary:
1. `cpp`
2. `csharp`
3. `java`
4. `js`
5. `kotlin`
6. `objc`
7. `php`
8. `pyi`
9. `python`
10. `ruby`

Normally for the above plugins, `buf` will execute the `protoc` binary that is found in
your `PATH`. But this configuration option lets you point to a specific binary. This
is particularly useful if you need to support a specific version of `protoc`, which
could differ from the version in `PATH`.
```

docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -201,6 +201,20 @@ $ protoc -I . $(find . -name '*.proto')
...
+
+```yaml title="buf.gen.yaml"
+version: v1
+plugins:
+  - plugin: go
+    revision: 1
+    out: gen/proto/go
+    protoc_path: /path/to/old-version
```
jhump, line 215:
```suggestion
    protoc_path: /path/to/specific/version/bin/protoc
```

jhump (commented):
BTW, @bufdev, `buf` currently includes `js` in the list to proxy to `protoc` (so I included it in this doc). However, as of v21 (from [May 2022](https://developers.google.com/protocol-buffers/docs/news/2022-05-06)), `js` is no longer part of `protoc`.

docs/configuration/v1/buf-gen-yaml.md
jhump: @bufdev, I took a stab at it. WDYT?

bufdev:
> BTW, @bufdev, buf currently includes js in the [list to proxy to protoc](https://github.com/bufbuild/buf/blob/main/private/pkg/app/appproto/appprotoexec/appprotoexec.go#L49-L62) (so I included it in this doc). However, as of v21 (from [May 2022](https://developers.google.com/protocol-buffers/docs/news/2022-05-06)), js is no longer part of protoc.

We're going to have to do some custom logic inside https://github.com/bufbuild/buf/blob/main/private/pkg/app/appproto/appprotoexec/protoc_proxy_handler.go for this, will file an issue on bufbuild/buf

docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -201,6 +201,38 @@ $ protoc -I . $(find . -name '*.proto')
...
+#### `protoc_path`
+The `protoc_path` of a plugin is optional and only applies to the code generators
+that are _built-in_ to `protoc`. Normally, a plugin is a separate executable whose
+binary name is like `protoc-gen-<name>`. But for a handful of plugins, the executable
+used is `protoc` itself. The following plugins result in invoking `protoc` instead of
+a dedicated plugin binary:
+1. `cpp`
+2. `csharp`
+3. `java`
+4. `js`
```
bufdev, line 213: Say for <v21

docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -201,6 +201,38 @@ $ protoc -I . $(find . -name '*.proto')
...
+5. `kotlin`
```
bufdev, line 214: Say for >=3.17

docs/configuration/v1/buf-gen-yaml.md
```diff
@@ -201,6 +201,38 @@ $ protoc -I . $(find . -name '*.proto')
...
+6. `objc`
+7. `php`
+8. `pyi`
+9. `python`
+10. `ruby`
+
+Normally for the above plugins, `buf` will execute the `protoc` binary that is found in
+your `PATH`. But this configuration option lets you point to a specific binary. This
```
bufdev, line 222: $PATH

bufdev (approved):
LGTM pending comments being addressed. Merge after the feature is merged

tonyli233:
Updated the changes. I just want to make sure that's correct.

PR #368, opened 2023-02-08
Add nodes to document tree

elliotmjackson:
Pathing the way for upcoming getting started guides

1. Create new `Guides` category to house tour and how-to.
2. Create new `The Buf CLI` category to include cli features

no links change, only sidebar display.


sidebars.js
```diff
@@ -6,86 +6,112 @@ const sidebars = {
...
-                    customProps: {
-                        badge: {
-                            label: "new",
-                            severity: "info"
-                        }
-                    }
+                    type: "category",
+                    label: "Tour",
+                    items: [
+                        "tour/introduction",
```
joshcarp, line 15: Question: Would it be possible to change the `introduction` to `overview`. This way there's a nice symmetry between this and the buf cli reference docs which has `overview` as the category index page?

elliotmjackson: out of scope, the tour is getting deleted

joshcarp:
Do we need to make it so that old links aren't broken?

elliotmjackson:
highlighted in the description of the PR "no links change, only sidebar display".

joshcarp (approved):

PR #380, opened 2023-02-13
Add Language-specific guides

elliotmjackson:


docs/guides/language/go/05_configure-cicd.md
```diff
@@ -0,0 +1,26 @@
+---
+sidebar_position: 5
+id: configure-ci-cd
+title: Configure CI/CD
```
fyockm, line 4: why does this belong in the language specific guide? feels language agnostic to me.

elliotmjackson: [this](https://github.com/bufbuild/docs.buf.build/pull/381) is more worthwhile

PR #383, opened 2023-02-14
Add make updatereference that updates cli reference docs

joshcarp:
Adds `make  updatereference` that will update https://docs.buf.build/reference/cli/buf

This will be run on release on the next release and is compatible with anything after `v1.14.0` because `buf webpages` was only introduced after then


Makefile
```diff
@@ -60,3 +60,11 @@ endif
 	$(SED_I) "s/version = \"v[0-9].[0-9][0-9]*\.[0-9][0-9]*\"/version = \"v$(VERSION)\"/g" docs/build-systems/bazel.md
 	$(SED_I) "s/BUF_VERSION=[0-9].[0-9][0-9]*\.[0-9][0-9]*/BUF_VERSION=$(VERSION)/g" docs/ci-cd/setup.mdx
 	$(SED_I) "s/downloadRelease: \"[0-9].[0-9][0-9]*\.[0-9][0-9]*\"/downloadRelease: \"$(VERSION)\"/g" docusaurus.config.js
+
+.PHONY: updatereference
+updatereference:
+ifndef VERSION
+	$(error "VERSION must be set")
+endif
+	go install github.com/bufbuild/buf/cmd/buf@$(VERSION); \
```
elliotmjackson, line 69:
```suggestion
	go install github.com/bufbuild/buf/cmd/buf@v$(VERSION); \
```

elliotmjackson (approved):

PR #396, opened 2023-02-18
Rewrite breaking overview and rules

twilly:
The description of the different categories was a little thick and not clear why one would choose a particular breaking category. This is my attempt at making breaking documentation a tad friendlier.

Overview is much simpler intro and asks the reader to go to the rules page for more details.

The order of categories is now always presented in strict to least strict (FILE -> PACKAGE -> WIRE_JSON -> WIRE).

Bullet points are shorter and to the point. More descriptive details are in two sections: FILE and PACKAGE, WIRE and WIRE_JSON. These sections are chosen because it groups two higher level categories: generated code breakage versus encoding breakage.


bufdev (commented):
My read is this might be over correcting in the opposite direction, both providing too little context in the summary and perhaps too much in the explanation. Let's sync about it on Tuesday

bufdev (approved):

bufdev:
Talked about this, we're good.

PR #442, opened 2023-03-20
Update page on authentication

joshcarp:
The internal discussion was that some users had issues and set BUF_TOKEN locally which was an invalid token and `buf logout` wasn't doing anything. This reorders the docs and makes it clear that the preferred way of authenticating locally is with `buf registry login` instead of BUF_TOKEN. Also adds some links to the buf cli reference docs for `buf registry login` and `buf registry logout`


rhbuf:
@joshcarp I left some comments that are not related to your changes, but since we're here I think we can increase readability on this page

rhbuf (commented):
@joshcarp I left some comments that are not related to your changes, but since we're here I think we can increase readability on this page

